PR #5, opened 2021-06-29
Feature/118/Create-Issue

MarcusviniciusLsantos:
Create issue screen:  
image


MarcusviniciusLsantos:
## Current screen: 

- [ ] added max button functionality
- [ ] added title and description validation

image

pages/create-issue.tsx
```diff
@@ -45,14 +47,18 @@ export default function PageCreateIssue() {
 
   }
 
+  const title_and_description_validation = () => {
+    return (issueTitle.length && issueDescription.length) ? false : true
```
moshmage, line 51: This is redundant as `issueTitle.length && issueDescription.length` already returns a `boolean`

```
return !(issueTitle.length && issueDescription.length)
```

pages/create-issue.tsx
```diff
@@ -1,14 +1,16 @@
...
 import BeproService from '../services/bepro';
 import GithubMicroService from '../services/github-microservice';
 
 export default function PageCreateIssue() {
 
-
   const [issueTitle, setIssueTitle] = useState("");
   const [issueDescription, setIssueDescription] = useState("");
-  const [issueAmount, setIssueAmount] = useState<number>(0);
+  const [issueAmount, setIssueAmount] = useState<string>('0');
```
moshmage, line 12: `issueAmount` can't be a string because `balance` isn't a string. When comparing the two we will need to make type casting when we shouldn't need to. These 2 values must be of the same _type_

pages/create-issue.tsx
```diff
@@ -45,14 +47,18 @@ export default function PageCreateIssue() {
 
   }
 
+  const title_and_description_validation = () => {
```
moshmage, line 50: please, follow naming conventions of with cammelCase: `titleAndDescriptionValidation` but that's not really what this function does, so `titleAndIssueExist` should be more telling of what this function _actually_ does

moshmage (changes requested):

helpers/number-transformation.ts
```diff
@@ -0,0 +1,7 @@
+export const toNumber = (v: number|string) => +v.toString().replace(',', '').replace(/\D+/g, '.');
```
moshmage, line 1: We can replace the return value from this function to,
```
new Intl.NumberFormat('en-gb').format(`123456`.replace(/\D+|,/, ''))
```

pages/create-issue.tsx
MarcusviniciusLsantos: done

pages/create-issue.tsx
MarcusviniciusLsantos: Ok, done.

pages/create-issue.tsx
MarcusviniciusLsantos: here I have a problem. Because status as number does not allow writing "," and "."

pages/create-issue.tsx
```diff
@@ -45,14 +61,50 @@ export default function PageCreateIssue() {
...
+  const titleAndIssueExist = () => {
+    return !(issueTitle.length && issueDescription.length)
+  }
+
+  const verifyAmountBiggerThanBalance = () => {
+    return !(issueAmount.floatValue > Number(balance))
+  }
+
+  const checksToEnableCreateIssue = () => {
+    if (!allowedTransaction){
```
moshmage, line 73: you don't need to `else if` if you have returned from the statement,

```ts
checksToEnableCreateIssue = () => {
  if (!allowedTransaction)
    return true
    
  if (titleAndIssueExist())
    return true

  // ...
}
```

But then again, the naming of this function is all over the place, _and_ this is used on a "disabled" prop meaning the person reading this code will have to know, beforehand, that this function does not do what it says it does.
And you can also simplify this function to its purest forms:

We know that all of these checks need to be true, otherwise the button will be disabled. meaning that if any of these values is false, we return true:

```ts
function isButtonDisabled(): boolean {
  return [
      allowedTransaction,
      titleAndIssueExist(),
      verifyAmountBiggerThanBalance(),
      issueAmount.floatValue > 0 || !!issueAmount.formattedValue
    ].some(v === false);
}
```

**NOTICE** for some reason, all the functions are in the positive, but their return value is negative;
Take this function for instance: "title and issue exist" meaning that If I have a title and a description this function should return true but it doesn't:

|combination|result|
|---|---|
|!(1 && 1)|false|
|!(1 && 0)|true|
|!(0 && 1)|true|
|!(0 && 0)|true|

```ts
  const titleAndIssueExist = () => {
    return !(issueTitle.length && issueDescription.length)
  }
 ```
 
to make it easy on everyone I'd simply rewrite all the functions to be _affirmitive_ and return _true_:

```
const issueContentIsValid = () =>  !!issueTitle && !!issueDescription;
```

And then I would do the same for every "is X valid" function

Comment (MarcusviniciusLsantos) [reply pages/create-issue.tsx:73]: Good. Adjustments done!

File: styles/styles.scss [hunk 6]
```diff
@@ -339,6 +339,22 @@ p {
...
     color: $gray;
   }
+
+  .input-group {
+    .input-group-text{
+      background-color: $black;
+      border: 1px solid $black;
+      color: $gray;
+    }
+    input[type=number]::-webkit-inner-spin-button,
```
moshmage, line 349: @andredbsantos While I also don't like the "up/down" arrows on a input, and while I didn't spot any on the design, should we _hide_ these by default every time ? (just making sure since this _is_ a design choice)

pages/create-issue.tsx
```diff
@@ -61,45 +105,54 @@ export default function PageCreateIssue() {
...
                   </div>
-                  <div className="form-group mb-4">
-                    <label className="p-small trans mb-2">Set $BEPRO ammout</label>
-                    <input min="0" type="number" className="form-control" placeholder="0"
-                      value={issueAmount}
-                      onChange={e => setIssueAmount(parseInt(e.target.value))}/>
+                  <div className="form-group col-md-4 mb-4">
+                    <label className="p-small mb-2">Set $BEPRO value</label>
+                    <div className="input-group">
+                      <NumberFormat min="0" max={`${balance}`} className="form-control" placeholder="0"
```
soujvnunes, line 123:
```suggestion
                      <NumberFormat min="0" max={balance} className="form-control" placeholder="0"
```
You can remove the template on `[max]` and set the expression directly.

pages/create-issue.tsx
```diff
@@ -61,45 +105,54 @@ export default function PageCreateIssue() {
...
+                        value={issueAmount.formattedValue}
+                        thousandSeparator={true}
+                        onValueChange={handleIssueAmountOnValueChange}
+                        onBlur={handleIssueAmountBlurChange}/>
+                      <span className="input-group-text text-white-50 p-small">$BEPRO</span>
+                    </div>
```
soujvnunes, line 129:
```suggestion
                    <InputNumber {...props} />
```
I think this composition for the input number is pretty interesting. Would you create a component, like `InputNumber´ and some resources used 'outside' to be within, so we can use it elsewhere?

pages/create-issue.tsx
```diff
@@ -1,23 +1,39 @@
...
-  const [issueAmount, setIssueAmount] = useState<number>(0);
+  const [issueAmount, setIssueAmount] = useState<Amount>({ value: '0', formattedValue: '0', floatValue: 0});
+  const [balance, setBalance] = useState<string>('0');
   const [allowedTransaction, setAllowedTransaction] = useState<boolean>(false);
 
+
+  useEffect(() => {
+    const resBepro = async () => {
+      await BeproService.login();
+      setBalance(await BeproService.network.getBEPROStaked())
```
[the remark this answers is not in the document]
MarcusviniciusLsantos: Good question. @moshmage "Is this `balance` state supposed to change at anytime, besides to catch it on component mount?"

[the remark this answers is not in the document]
MarcusviniciusLsantos: Create a helper for get "beprobalance" -> c7d5f37

pages/create-issue.tsx
MarcusviniciusLsantos: Lack of attention from me, sorry. Done -> dfb16bd

pages/create-issue.tsx
MarcusviniciusLsantos: Good idea. I will start working on it.

pages/create-issue.tsx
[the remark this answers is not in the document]
moshmage: Not needed, bepro service is already as the service itself works as a "helper" (as in: we gain nothing by refactoring that part into a function)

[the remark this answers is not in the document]
soujvnunes: Can you clarify @moshmage where else, from a top-level, can we call `login()`? Because aparently it has to be called before every BeproService module for this to work. If this can't happen, we must create some custom hook, as I suggested, so `login()` will be already within and called just one time at one place.

pages/create-issue.tsx
MarcusviniciusLsantos: Done -> 2ecd986

pages/create-issue.tsx
```diff
@@ -1,48 +1,101 @@
...
-    const beproAddress = await BeproService.getAddress();
-    const payload = {
-      amount: issueAmount,
-      address: beproAddress
-    }
-    const res = await BeproService.network.approveTransactionalERC20Token();
-    const resApproved = await BeproService.network.isApprovedTransactionalToken(payload);
-    if (resApproved) {
+    evt.preventDefault();    
+    try {
```
moshmage, line 34: don't use try catch without logging the catch as that will not show on the console log as an error and you will throw future coders off scent if a bug is found here.

In general, avoid using try-catch formula for more than one statement, and if you really have to, then PLEASE log what broken on the catch.

pages/create-issue.tsx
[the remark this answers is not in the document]
moshmage: @soujvnunes we currently don't call it on a "top-level", as that's going to be handled by "My Account"; 

If the user needs a call to a `login()` statement somewhere before getting to this page, we should then create a navigation guard that checks for the logged in state _before_ the route is even shown - and if !login then we redirect the user to the "My Account" page where he can then connect/login against the metamask wallet and then be redirected back to where he was.


Again: A exported function for this sole propose is not needed nor should we have it as a dependency, we should use the service correctly instead, and use a navigational guard, and eventually account with the state from My Account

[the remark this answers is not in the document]
soujvnunes: Now I get it! @moshmage thank you!

moshmage (changes requested):
Comments are spread on the issue because stuff got outdated before I could post all the comments

pages/create-issue.tsx
```diff
@@ -61,45 +114,51 @@ export default function PageCreateIssue() {
...
                       />
+                    <p className="p-small trans my-2">Tip: Try to be as much descriptive as possible</p>
                   </div>
-                  <div className="form-group mb-4">
-                    <label className="p-small trans mb-2">Set $BEPRO ammout</label>
-                    <input min="0" type="number" className="form-control" placeholder="0"
-                      value={issueAmount}
-                      onChange={e => setIssueAmount(parseInt(e.target.value))}/>
+                  <div className="form-group col-md-4 mb-4">
+                    <label className="p-small mb-2">Set $BEPRO value</label>
```
soujvnunes, line 130: If you take a look in general for all inputs, this label could be within `InputNumber` too! So, it could render the label through `[label]`.

If other props has a value per default, you can set it directly within, such `thousandSeparator` if so.

components/inputNumber.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export default function InputNumber({
+  children = "$BEPRO",
+  placeholder = "0",
+  ...params
+}: NumberFormatProps) {
+  return (
+    <div className="input-group">
+      <NumberFormat {...params} placeholder={placeholder} />
+      <span className="input-group-text text-white-50 p-small">{children}</span>
```
soujvnunes, line 12: Apparently, this span render a thing called 'tick' or 'symbol' (related to Currency Symbol/**$BEPRO** in this case). Would you not find interesting to define it through `[symbol]` instead of `[children]`?

soujvnunes, line 12: Apparently, this span render a thing called 'tick' or 'symbol' (related to Currency Symbol/**$BEPRO** in this case). Would you not find interesting to define it through `[symbol]` instead of `[children]`?

pages/create-issue.tsx
```diff
@@ -1,48 +1,101 @@
...
+  const [balance, setBalance] = useState<string>('0');
   const [allowedTransaction, setAllowedTransaction] = useState<boolean>(false);
 
+  const useBeproBalance = async () => {
+    setBalance(await BeproBalance())
+  }
+
+  useEffect(() => {
+    useBeproBalance()
+  },[])
```
soujvnunes, line 29: You can set the `balance` state updater directly as a side-effect. Like:

```jsx
 useEffect(() => {
    setBalance(BeproBalance())
  }, [BeproBalance])
```

In the current usage way, `useBeproBalance()` is getting a new reference on every render, getting rerendered too. To avoid this behavior, must memoize it (what is expensive), like:
```jsx
 const useBeproBalance = useCallback(async () => {
    setBalance(await BeproBalance())
 }, [])

useEffect(() => {
    useBeproBalance()
 }, [useBeproBalance])
```

pages/create-issue.tsx
MarcusviniciusLsantos: Done. -> [991a937, bd13aeb]. does this solve?

pages/create-issue.tsx
MarcusviniciusLsantos: Done. -> 741c2a7. does this solve?

components/inputNumber.tsx
MarcusviniciusLsantos: Done. -> 741c2a7. does this solve?

helpers/bepro-balance.ts
```diff
@@ -0,0 +1,8 @@
+import BeproService from '../services/bepro';
+
+const BeproBalance = async () => {
```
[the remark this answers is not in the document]
MarcusviniciusLsantos: I removed it as a suggestion from the mosh in the meeting

pages/create-issue.tsx
MarcusviniciusLsantos: Done. -> 98d6e6e. Does this solve?

soujvnunes: Solves! But was created an unnecessary expensive memoization, as `useBeproBalance()` it's been called just once and by callback hook. By now, in this case, we need to focus on get this BEPRO Staked balance and set it both on **create-issue mount** and when its own **value change**.  Let's just use the getter module inside the `useEffect()` (so we don't need to define this function as an dependency).

Instead of this:

```jsx
  const useBeproBalance = useCallback(async () => {
    await BeproService.login(); 
    setBalance(await BeproService.network.getBEPROStaked())
  }, [])

  useEffect(() => {
    useBeproBalance()
  },[useBeproBalance])
```

Let's do this:

```jsx
useEffect(() => {
  /* > As we're calling an async function, using this to conditionally update
  *  > component state prevents memory leak.
  */
  let isMounted = true;

  void async function getBeproBalance() {
    await BeproService.login(); 

    if (isMounted) {
      setBalance(BeproService.network.getBEPROStaked());
    }
  }();

  return () => {
    isMounted = false;
  }
}, [])
```
Let me know if makes sense for you.

pages/create-issue.tsx
soujvnunes: It does! 😍 I just would remove this `[symbol]` default value, because in the future we may use this `InputNumber` for other purposes besides entry balance value. Don't you agree? So:

```jsx
  symbol = "",
```
```jsx
  {symbol && <span>{symbol}</span>}
```

helpers/bepro-balance.ts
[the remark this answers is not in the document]
soujvnunes: Yup!

pages/create-issue.tsx
MarcusviniciusLsantos: Good. The `isMounted` makes it wait for the component to actually mount. behaves like conventional react componentDidMount, right?

MarcusviniciusLsantos: done -> e16601a

pages/create-issue.tsx
moshmage: I knew I had lost something in translation :D -- we'll talk on the standup 👍

pages/create-issue.tsx
```diff
@@ -1,48 +1,122 @@
...
+  const [issueAmount, setIssueAmount] = useState<Amount>({ value: '0', formattedValue: '0', floatValue: 0});
+  const [balance, setBalance] = useState<string>('0');
   const [allowedTransaction, setAllowedTransaction] = useState<boolean>(false);
+  const router = useRouter()
+
+  useEffect(() => {
+    let isMounted = true;
+    void async function getBeproBalance() {
+      await BeproService.login(); 
+      if (isMounted) {
```
moshmage, line 27: wouldn't this _always be true_ ? and wouldn't that make it so we're calling it on every update?

soujvnunes: This to prevent calling the state updater when the components unmount!

pages/create-issue.tsx
soujvnunes: More or less. Complementing, we allow updates trigger only on mounted components. `isMounted` don't stop the mouting/unmounting of a component. This is just a common variable that helps us to handle out-of-order fetch calls. In class components, would be something like:
```jsx
constructor() {
  this.isMounted = false;
}

componentDidMount() {
  this.isMounted = true;

  if (this.isMounted) {
    // do something
  }
}

componentWillUnmount() {
  this.isMounted = false;
}

**Review by moshmage — CHANGES_REQUESTED** [review]

File: pages/create-issue.tsx (hunk 15 above)
Comment (soujvnunes) [reply pages/create-issue.tsx:27]: As we are using an async promise call, we must use a reference variable to check already unmounted component for the next treatment of async response (avoiding memory leaks). When settled on an async function as a side-effect, this will happen on our [console debug](https://reactjs.org/docs/hooks-rules.html#eslint-plugin) when the component umount:
```console
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```
Why? The memory leak will happen if the microservice took time to respond and the component was unmounted before the response was received. Though the component was unmounted, the response to the request will still be received on completion. And guess what will be triggered even when the component has been unmounted?
```jsx
// based on the suggestion you gave
useEffect(() => {
  async function getBalance() {
    await BeproService.login();
    return BeproService.network.getBEPROStaked();
  }
  
  // 👇 will be triggered even when the component has been unmounted.
  // Besides, updater state it's naturally async. I don't think use await here is necessary
  setBalance(await getBalance())
  
  return () => null;
}, [])
```
That's where I accomplished on the early suggestion. Let's wrapped it up:
```jsx
useEffect(() => {
  /* As we're calling an async function, using this to conditionally update
  *  component state prevents memory leak.
  */

  /* Everyhing here will be triggered only
  * on component mount. In this case,
  * we use a variable to reference when
  * the component mounts/umounts.
  * */
  let isMounted = true;

  void async function getBeproBalance() {
    await BeproService.login(); 

    // this will be only triggered if the component is mounted, free of wait for responses.
    if (isMounted) {
      setBalance(BeproService.network.getBEPROStaked());
    }
  }();

  /* Everyhing here will be triggered only
  * on component unmount. Set null here,
  * will just trigger nothing. In this case, we
  * set the reference as 'umount' (isMounted = false)
  * */
  return () => {
    isMounted = false;
  }
}, [])
```

Related: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

pages/create-issue.tsx
```diff
@@ -1,48 +1,122 @@
...
     }
     const beproAddress = await BeproService.getAddress();
-    const contractPayload = {tokenAmount: issueAmount, cid: beproAddress};
-    const res = await BeproService.network.openIssue(contractPayload);
-    console.log("🚀 ~ file: create-issue.tsx ~ line 41 ~ createIssue ~ res", res)
+    const contractPayload = {tokenAmount: issueAmount.floatValue, cid: beproAddress};
+    await BeproService.network.openIssue(contractPayload)
+      .then(async (response) => {
+        payload.issueId = response.events?.OpenIssue?.returnValues?.id;
+        await GithubMicroService.createIssue(payload)
```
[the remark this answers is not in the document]
MarcusviniciusLsantos: Done -> bd23031

pages/create-issue.tsx
```diff
@@ -34,25 +34,22 @@ export default function PageCreateIssue() {
...
-      .approveTransactionalERC20Token()
-      .then(async () => {
-        await BeproService.network
-        .isApprovedTransactionalToken(payload)
-        .then(() => {
-          setAllowedTransaction(true);
-          setLoadingAttributes(false);
-        })
-        .catch((error) => handleErrorAndLoading(error))
+      .then(() => { beproAddress = BeproService.getAddress() })
```
moshmage: turn this around, do approveTransactionaERC20Token() before .getAddress():

```ts
setLoadingAttributes(true) // BEFORE ANYTHING SET LOADER TO TRUE
await BepropService.login()
  .then(BeproService.network.approveTransactionalERC20Token) // approve transaction
  .then(BeproService.getAddress) // get and return the address 
  .then(address => BeproService.network.isApprovedTransactionalToken({address, amount: issueAmount.floatValue})) // you don't need a const for a value that you'll use right after, just provide the payload as an argument instead of as a const
  .then(transaction => {
    setAllowedTransaction(transaction)
    setLoadingAttributes(false)
  })
  .catch((error) => handleErrorAndLoading(error));

File: pages/create-issue.tsx [hunk 18]
```diff
@@ -68,9 +65,9 @@ export default function PageCreateIssue() {
     const beproAddress = await BeproService.getAddress();
     const contractPayload = {tokenAmount: issueAmount.floatValue, cid: beproAddress};
     await BeproService.network.openIssue(contractPayload)
-      .then(async (response) => {
+      .then((response) => {
```
moshmage, line 68: This then can only have _one catch_

moshmage (changes requested):

pages/create-issue.tsx
moshmage: image

https://user-images.githubusercontent.com/1077079/125058621-5d535c00-e0a2-11eb-8bfd-05202b8f7fe6.mp4

pages/create-issue.tsx
MarcusviniciusLsantos: done

moshmage (approved):

moshmage (approved):

moshmage (approved):

moshmage (approved):

PR #7, opened 2021-06-29
main - issue list

andrefcgoncalves:


pages/account.tsx
```diff
@@ -2,26 +2,31 @@ import { GetStaticProps } from 'next'
...
   useEffect(() => {
     getMyIssues();
   }, []); // initial load
 
   const getMyIssues = async () => {
     await BeproService.login();
     const beproAddress = await BeproService.getAddress();
-    setMyIssues(await BeproService.network.getIssuesByAddress(beproAddress));
+    let issueIds = await BeproService.network.getIssuesByAddress(beproAddress);
+    issueIds = issueIds.map((i) => i+1)
```
moshmage, line 20: This is odd, why do we need to sum `1` to the issueId?

services/github-microservice.ts
```diff
@@ -6,6 +6,9 @@ export default class GithubMicroService {
 
   static async createIssue(payload) {
     await axios.post(API_HOST  + '/issues', payload);
-
+  }
+  static async getIssues(issueIds) {
```
moshmage, line 10: What's the type of `issueIds`?

moshmage (changes requested):

pages/account.tsx
```diff
@@ -33,12 +38,12 @@ export default function PageAccount() {
 
         <div className="container">
           <div className="row justify-content-center">
-            <div className="col-md-10">
-              <IssueListItem></IssueListItem>
-            </div>
-            <div className="col-md-10">
-              <IssueListItem></IssueListItem>
-            </div>
+            {myIssues.map((issue) => (
```
moshmage, line 41: can we make a component out of this? from experience, writing this small loop will get old pretty quick. And we will start making it differently because we will get tired of writing it.

moshmage (approved):

PR #8, opened 2021-07-02
Feature/155/global loading state

MarcusviniciusLsantos:
- [x] create loading component
- [x] adding to global scope

- ## Video showing Loading component in a controlled environment

https://user-images.githubusercontent.com/34746557/124202835-afe4b500-dab1-11eb-936a-56607a56bae1.mp4

https://user-images.githubusercontent.com/34746557/124202845-b5da9600-dab1-11eb-96db-cb04009b8a30.mp4


pages/_app.tsx
```diff
@@ -4,19 +4,24 @@ import { AppProps } from 'next/app'
...
-  return <>
+
+  return ( 
+  <>
     <Head>
     </Head>
     <MainNav></MainNav>
-    <Component {...pageProps} />
-
+    <Component loading={setLoading} {...pageProps} />
```
moshmage, line 23: It'd be cleaner if `Loading` component exported a context of its own and then we could simply use it on all the application without having to connect `Component` to `Loading` since `Loading` would be using its own context that would be used by every other component that needs to show the loading

MarcusviniciusLsantos: Done

MarcusviniciusLsantos:
## Example usage in code
`const loading = useContext(LoadingContext)`
`loading(true)`
`loading(false)`

providers/loading.tsx
```diff
@@ -0,0 +1,19 @@
+import { createContext, Dispatch, SetStateAction, useState } from "react";
+import Loading from "../components/loading";
+
+let setStateAction: Dispatch<SetStateAction<boolean>>;
+
+const LoadingContext = createContext(setStateAction);
+
+function LoadingContextProvider({ children }) {
+  const [loading, setLoading] = useState<boolean>(false);
```
moshmage, line 9: make this be global to this file, so we can export `loading` variable in case we might need to read the state,

```ts
// ...
const LoadingContext = createContext(setStateAction);
const [isLoading, setLoading] = useState<boolean>(false);

function LoadingContextProvider...
```

providers/loading.tsx
```diff
@@ -0,0 +1,19 @@
+import { createContext, Dispatch, SetStateAction, useState } from "react";
+import Loading from "../components/loading";
+
+let setStateAction: Dispatch<SetStateAction<boolean>>;
+
+const LoadingContext = createContext(setStateAction);
```
moshmage, line 6: `setStateAction` is being used as a default value, but the default value typed does not match the value that LoadingContext will have, nor is the variable initialized with it - so you're initializing this state with an undefined default state, `createContext(false)` is more than enough

providers/loading.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  return (
+    <LoadingContext.Provider value={setLoading}>
+      <Loading show={loading} />
+      {children}
+    </LoadingContext.Provider>
+  );
+}
+
+export { LoadingContextProvider, LoadingContext };
```
moshmage, line 19: You can be more expressive in this return statement, so consumers know whats what:

```ts
export { LoadingContextProvider, isLoading, setLoading: LoadingContext }
```

This would then be consumed (on other components other than app)

```ts
import {isLoading, setLoading} from '../providers/loading-provider';

setLoading(true) // shows
console.asset(isLoading === true, `isLoading value should be true`)
setLoading(false); // hides
console.assert(isLoading === false, `isLoading value should be false`)
```

moshmage (changes requested):

providers/loading.tsx
MarcusviniciusLsantos: `export { LoadingContextProvider, isLoading, setLoading: LoadingContext }`

have a problem `setLoading: LoadingContext`.

- solution suggestion 

> `const [isloading, setLoading] = useState<boolean>(false);`
> `const LoadingContext = createContext(isloading);`

> `export { LoadingContextProvider, isloading, setLoading: LoadingContext };`


Can be?

providers/loading.tsx
MarcusviniciusLsantos: done

MarcusviniciusLsantos: image

moshmage: yeah, I should have tested that before :D

lets go the cheeky route:

```ts
//...
let getLoadingState: () => boolean;
// ...
function LoadingContextProvider {
    const [loading, setLoading] = useState<boolean>(false);
    getLoadingState = () => loading;
    // ...
}

export { LoadingContextProvider, LoadingContext, getLoadingState };

Comment (MarcusviniciusLsantos) [reply providers/loading.tsx:9]: Good, i adjusted. Done ^.^

**Review by moshmage — APPROVED** [review]

File: providers/loading.tsx (hunk 2 above)
Comment (soujvnunes) [reply providers/loading.tsx]: > use state only works within a component

You always call hooks within CamelCase components or modules, as helpers itself. Example:
```
function Component() {
   const [S, D] = useState()
}
```

and not as
```
function component() {
   const [S, D] = useState()
}
```

PR #9, opened 2021-07-02
Feature/175

soujvnunes:


helpers/isWebThreeInstalled.ts
```diff
@@ -0,0 +1,13 @@
+declare global {
+  interface Window {
+    ethereum: any;
+  }
+}
+
+export default function isWebThreeInstalled() {
```
moshmage, line 7: you can summarize this into
```ts
export default function isWebThreeInstalled(): boolean { return !!window?.ethereum }

File: plain-text.ts [hunk 2]
```diff
@@ -0,0 +1,10 @@
+export default {
```
moshmage, line 1: We need to select a i18n plugin and then start doing this but for now just add this text onto the component itself

helpers/index.ts
```diff
@@ -0,0 +1 @@
+export { default as isWebThreeInstalled } from "./isWebThreeInstalled";
```
moshmage, line 1: Don't make this export barrel or import lines will become pretty large when we more than a few helpers

moshmage (changes requested):

helpers/isWebThreeInstalled.ts
soujvnunes: I can't. Since `window?` equals to `window && window` and besides it came bool, could be `undefined` too, throwing the error in this case (bc it's using an undefined variable for verification).

plain-text.ts
soujvnunes: Ok!

helpers/index.ts
soujvnunes: I do. That's exactly the point when dealing with more than one helper module. This solves the problem on importing default named modules at components. Instead of this:
```jsx
import module1 from "helpers/module1";
import module2 from "helpers/module2";
import module3 from "helpers/module3"
```

the current structure enhances import calls, like:
```jsx
import { module1, module2, module3 } from "helpers";
```

moshmage: @soujvnunes that import helpers line will get big in a matter of seconds, besides, the only function that exists is `isWebThreeInstalled`, no point making an export barrel for one function. But even so, you exported that function as a `default` meaning _you want_ people to consume it as 

```
import function from `file.ts`
```

But then you went and _import the deconstruction of the module_ only to export that as isWebThreeInstalled again, when all you had to do was:

```
import isWebThreeInstalled from `@helpers/is-web-3-installed.ts
```

And all would be cool with everything because the person reading it didn't have to jump through two files to get to the source of the file.

helpers/isWebThreeInstalled.ts
moshmage: so: `window` will NEVER be udefined. No need to check for that, the null coallescent was there just so you could translate what I wrote for what you had. also, this function _has_ to return a boolean, you said so yourself here: https://github.com/bepronetwork/web-network/pull/9/files#diff-dd2381d8e5f35d95030468f5399bf2e507165edb616304f26aff894ccd5b013aR9

and while `null` is falsy, it's not a boolean. null is null. but since `window` will never be undefined, `return null` will never be read - so you dont need that there.

And since we know window will never be undefined, you also don't need the `if()`

leaving the fuction with `!!window.etherium` as its true form

moshmage (changes requested):

helpers/isWebThreeInstalled.ts
soujvnunes: I'm sorry, I should have shown the recipe before:

image

Again, the summarized suggestion you provided do not work in this case. I return `null` for where we're using this modules do not receive `undefined` as well, when jump of through the `window` check,

helpers/index.ts
soujvnunes: Yup! And I think we will not do this in most components besides where currently it is. I'm gonna remove the indexation

moshmage (approved):

moshmage (approved):

moshmage (approved):

moshmage (approved):

PR #11, opened 2021-07-08
(feat 128) github handle component

moshmage:
https://app.clubhouse.io/bepro/story/128/github-connect-with-account

https://app.clubhouse.io/bepro/story/249/bugs-urgent

PR #14, opened 2021-07-18
Adjustments in the loading lifecycle

MarcusviniciusLsantos:
- Adjusting the global loading component and creating a local loading component

This helps API requests like issues listings


moshmage:
Lets wait for #11 since it introduces a application state and we can then use that to the loaders advantage

PR #17, opened 2021-08-04
Feature/integrated methods issue-page

MarcusviniciusLsantos:


services/bepro-service.ts
```diff
@@ -12,7 +12,7 @@ class BeproFacet {
 
   constructor(
       public readonly web3Connection = `https://kovan.infura.io/v3/811fe4fa5c4b41cb9b92f9656aaeaa3b`,
-      public readonly contractAddress = `0x0CCb54d3b4248FE833aCaB4E8993Bc720737F4B3`,) {
+      public readonly contractAddress = `0xDf956E90a6E1c2BE14FB0c694389a0071EB6Eb91`,) {
```
moshmage, line 15: Choose the contract that's most up-to-date -- mine was an old one

services/bepro.ts
```diff
@@ -27,7 +27,7 @@ export default class BeproService {
     });
     // this.bepro.start();
     this.network = new Network({
-      contractAddress: `0x67d98dc9a353b72493c987d4d9abaf5114271675`,
+      contractAddress: "0xDf956E90a6E1c2BE14FB0c694389a0071EB6Eb91",
```
moshmage, line 30: This service will disappear, as in: we will not use it, and will delete it by next monday but I wouldn't mind having the same contract address you chose to `bepro-service` be the same

PR #33, opened 2021-08-07
Refactor: Filters by issue state

WellingtonMatheus:


pages/developers.tsx
```diff
@@ -37,49 +49,43 @@ const options_time = [
...
+    }
+    getIssues();
+  }, []);
+
+  const isDraftIssue = (issue: IssueData) => issue.state === 'draft';
+  const isClosedIssue = (issue: IssueData) => issue.state === 'closed';
+  const isOpenIssue = (issue: IssueData) => !isDraftIssue(issue) && !isClosedIssue(issue);
 
-  useEffect(getIssues, []);
+  const issuesFilteredByState = issues.filter(issue => {
```
moshmage, line 82: Please, use a function statement instead of a fat arrow constant to define functions. 

PS: _this_ is the place you could have written those constants as closures inside this function as it's not used anywhere else not would it hamper the readability of the code -- in fact: it would signal to the next developer that this 3 functions are only used inside that function.

pages/developers.tsx
```diff
@@ -37,49 +49,43 @@ const options_time = [
...
-    GithubMicroService.getIssues()
-                      .then(updateIssuesList)
-                      .catch((error) => {
-                        console.log('Error', error)
-                      })
-                      .finally(() => {
-                        dispatch(changeLoadState(false))
-                      });
-  }
+  useEffect(() => {
```
moshmage, line 63: Please, don't hide async effects inside a sync effect function with a closure. useEffects priority is too high to have its functions hidden away in a closure.

_if you don't need to call that function anywhere else, then simply don't state a function and instead write it as direct callback_

But we rather have the function stated _outside_ of the effect and then used inside the effect callback; If nothing else, for readability and understanding sake

pages/developers.tsx
```diff
@@ -8,23 +8,35 @@ import {ApplicationContext} from '../contexts/application';
 import {changeLoadState} from '../contexts/reducers/change-load-state';
 import {IssueData} from '../interfaces/issue-data';
 
-const options_issue = [
+type Filter = {
+  label: string;
+  value: string;
```
moshmage, line 13: If you typed the Filter prop, type the possible values as well - because _while they ARE strings_ that string value is being used as an enum

pages/developers.tsx
```diff
@@ -108,14 +114,11 @@ export default function PageDevelopers() {
...
-          {filterStateIssues.issues.length === 0 && !loading ? (
+          <ListIssues listIssues={issuesFilteredByState} />
+          {issuesFilteredByState.length === 0 && !loading.isLoading ? (
             <div className="col-md-10">
               <h4>
-                {filterStateIssues.state !== "all"
-                  ? filterStateIssues.state
-                  : null}{" "}
-                issues not found
+                {`${filterByState.emptyState}`}
```
moshmage, line 121: smart! :D

pages/developers.tsx
WellingtonMatheus: I tend to follow the guidelines of Dan Abramov, who wrote useEffect.

https://overreacted.io/a-complete-guide-to-useeffect/#moving-functions-inside-effects

So maybe I need to check if there are any dependencies missing in the dependency array, but asynchronous functions can still be inside useEffect, can't they?

moshmage: > but asynchronous functions can still be inside useEffect, can't they?

And here is the crux of the thing. **YOU CAN**, but there is an error deep in useEffect that will throw an error if it catches a state update when the component is no longer mounted. While highly improbable, _you can_ catch that error.

I usually hit this error when I give in an async function as a parameter to `useEffect`, such as

```
async function test() { return Promise.resolve(1) }
useEffect(test, [])
```

This happens because `async function` returns a Promise, while `useEffect` is expecting a "unmount function" to be returned on the use effect -- thus I made it a rule to never provide `async function` inside (or anywhere near) a `useEffect` call.

Following the same rule, I'm going "against it" if I "hide" an async call inside the useEffect. While it _should not_ throw the error because it _does run_ on the correct scope, _readers_ might think otherwise - and I've been favouring readability over code-golf lately.

PR #37, opened 2021-08-09
Fix: Format ETH and BEPRO balances

WellingtonMatheus:


moshmage (approved):

components/main-nav.tsx
```diff
@@ -93,7 +94,7 @@ export default function MainNav() {
             <Link href="/account" passHref>
               <a className="btn btn-md btn-trans mr-1">
                 <i className="ico-bepro mr-1"></i>
-                {beproBalance}
+                {formatNumberToNScale(beproBalance)}
```
moshmage, line 97: I think it would make sense to make this conversion on the BeproService side, that way we don't need to worry about that everywhere ?

PR #38, opened 2021-08-09
Hotfix/ALL TO DOs

MarcusviniciusLsantos:
- [ ] adding url from github when there are comments
- [ ] Should be equal to figma
- [ ] Should only show the Issues where the creator
- [ ] Should only show the Issues where the creator
- [ ] Should open the github page of the issue where the user can comment there directly
- [ ] If User isn´t logged/registerd with github the button should be disabled
- [ ] BEPRO amount image is cut, should be full
- [ ] ETH Amount should have a "," every 3 numbers and only 4 decimals (example : 330.2300 ETH)


helpers/formatBEPROS.ts
```diff
@@ -0,0 +1,12 @@
+export const formatBEPROS = (value: number) => {
```
moshmage, line 1: check #37

PR #44, opened 2021-08-11
fix: filter state issues

clarkjoao:
- fix query params to filter state
- create interface to `Issue State`


interfaces/issue-data.ts
```diff
@@ -1,3 +1,5 @@
+export type IssueState = 'ready' | 'draft' | null
```
moshmage, line 1: redeemed closed open

PR #47, opened 2021-08-11
Feature/proposal page

MarcusviniciusLsantos:
Adapting to new design and adding new logics 

image


components/issue-proposals.tsx
```diff
@@ -32,53 +34,167 @@ export default function IssueProposals({ numberProposals, issueId, amount }) {
...
+                            width: `${
+                              handlePercentage(
+                                toNumber(proposal.disputes),
+                                amount
+                              ) > 81.6 &&
+                              ((handlePercentage(
+                                toNumber(proposal.disputes),
+                                amount
+                              ) -
+                                81.7) *
```
moshmage, line 162: please, if these numbers are constants, set them as so at the top of the file.

styles/styles.scss
```diff
@@ -275,6 +275,12 @@ p {
   }
 }
 
+.btn-red {
```
moshmage, line 278: btn-danger exists, as well as border border-danger and color-danger

https://react-bootstrap.github.io/components/buttons/
https://getbootstrap.com/docs/4.0/utilities/borders/#border-color
https://getbootstrap.com/docs/4.0/utilities/colors/

styles/styles.scss
```diff
@@ -555,6 +575,12 @@ label {
 .bg-opac {
   background-color: rgba($white, 0.1);
 }
+.bg-purple {
+  background-color: $purple;
+}
+.bg-red {
```
moshmage, line 581: bg-danger exists

moshmage (changes requested):

components/issue-proposals.tsx
MarcusviniciusLsantos: ok, i adjusting

components/issue-proposals.tsx
```diff
@@ -46,11 +48,21 @@ export default function IssueProposals({ numberProposals, issueId, amount }) {
             merge_id: i,
           })
           .then((values: Proposal) => {
-            values.isDisputed = BeproService.network.IsMergeDisputed({
-              issueId: issueId,
-              mergeId: i,
-            });
-            arrayProposals.push(values);
+            BeproService.network
```
moshmage, line 51: you don't need to enter in a "callback hell", you can make these return their value and follow on the first then scope like so:

```ts
for (let i = 0; i < numProposals; i++) {
  const merge = await BeproService.network.getMergeById({issue_id, merge_id});
  const isMergeDisputed = await BeproService.network.IsMergeDisputed({issue_id, merge_id});
  const mergeProposal = await GithubMicroservice.getMergeProposalIssue({issue_id, merge_id});
  
  merge.isDisputed = isDisputed;
  merge.pullRequestId = mergeProposal.pullRequestId;
  
  arrayProposals.push(merge)
}
```

services/github-microservice.ts
```diff
@@ -126,4 +135,12 @@ export default class GithubMicroService {
                    return null;
                  })
   }
+  static async getMergeProposalIssue(issueId: string, MergeId: string) {
```
moshmage, line 138: note: this function can return null but the code never accounts for that, this will bring problems later on. account for this null return when you consume the value from this method.

moshmage (changes requested):

PR #51, opened 2021-08-12
fix: response erro to create pull request

clarkjoao:
- Add dalay to toast before remove
- Alert error to create PullRequest


components/toaster.tsx
```diff
@@ -11,18 +10,33 @@ export default function Toaster() {
     dispatch(removeToast(i));
   }
 
+  useEffect(() => {
```
moshmage, line 13: I forgot to add the `delay={3000}` prop to the `Toast` component. all we need to is add that prop and `onClose={() => dispatch(removeToast(i))}`

moshmage (approved):

moshmage (approved):

PR #56, opened 2021-08-17
Hotfix/GitHub login and create pull request

MarcusviniciusLsantos:
- [x] Adding GithubLogin where do you need it
- [x] Adding new Logic to signIn Github

### DIFF
--- a/services/github-microservice.ts
+++ b/services/github-microservice.ts
@@ -2,10 +2,10 @@ import axios from 'axios';
 import {IssueData, IssueState} from '@interfaces/issue-data';
 import { API } from '../env';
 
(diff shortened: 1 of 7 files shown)


services/github-microservice.ts
```diff
@@ -2,10 +2,10 @@ import axios from 'axios';
 import {IssueData, IssueState} from '@interfaces/issue-data';
 import { API } from '../env';
 
-interface User {
+export interface User {
```
moshmage, line 5: If you need to export an interface, export it from `@interfaces`

moshmage (approved):

PR #61, opened 2021-08-20
Feature/transaction popover

clarkjoao:
- Transaction Button
- Transaction List
- Transaction Modal Details


styles/styles.scss
```diff
@@ -671,6 +671,186 @@ label {
...
+  }
+  .popover-arrow{
+    &::after{
+      border-bottom-color: $bg-shadow;
+    }
+  }
+  
+
+  .transaction{
+    &__title {
```
moshmage, line 708: Please, don't use "body-like" CSS: follow the bootstrap implementation, as in: classes have styles, they can have relation, but they can't be bodies.

```
.transaction {
&__title {}
}
```

should read

```
.transaction {
 &.title {}
}
```

But I'm going to be honest and tell you that this looks like _loads_ of css for a simple modal. Either design doesn't follow correct bootstrap spacing (... it should) or we don't share the same spacing as the design (we should)

one of these is true, and we need to check which.


---

EDIT: be aware that you have access to the bootstrap contextual classes such as `text-white` and `border border-rounded border-info` and you might not need to add those in.

You also have `fs-*` classes (font size) that grow the font in relation to the base font size, as well as `mb-*` that apply bottom margins.

moshmage (changes requested):

moshmage:
Note: this component has mocked data; it needs to have _real_ data.

PR #63, opened 2021-08-20
Hotfix/button create proposal

MarcusviniciusLsantos:


components/create-proposal.tsx
```diff
@@ -73,10 +73,10 @@ export default function NewProposal({
         }).then(() => {
           handleClose();
           setDistrib({});
-        })
+        }).catch(() => setError("Error to create proposal in MicroService"))
```
moshmage, line 76: please, console.error the error so we can at least debug it when it happens

moshmage (changes requested):

moshmage (approved):

PR #82, opened 2021-09-06
Feature: nothing found redesign

WellingtonMatheus:


components/nothing-found.tsx
```diff
@@ -0,0 +1,24 @@
...
+  description: string;
+  action?: ReactNode;
+}
+
+export default function NothingFound({ description, action }: NothingFoundProps) {
+  return (
+      <div className="d-flex flex-column align-items-center gap-4">
+        <SearchIcon />
+        <p className="text-center text-uppercase fw-bold mb-0">{description}</p>
+        { action || null }
```
moshmage, line 15: use `children` instead, so we can write it as
```
<NothingFound description="thing">_whatever we want here_</NothingFound>
```
instead of passing html through an input parameter

moshmage (changes requested):

components/nothing-found.tsx
WellingtonMatheus: Done. Thanks!

moshmage (approved):

PR #83, opened 2021-09-08
fix: views bug 6th

clarkjoao:
- [x]  "Find issues to work" instead of "Find issue to work"
- [x]  Success color has to be changed to #35E0AD
- [x]  When picking a category in the issues states dropdown it shows a color that is not supposed to. Change this color to #1825AD


styles/styles.scss
```diff
@@ -646,6 +646,9 @@ label {
 .color-red {
   color: $red;
 }
+.color-green {
```
moshmage, line 649: change $success variable at variables scss and use success class

moshmage (changes requested):

pages/developers.tsx
```diff
@@ -101,6 +101,15 @@ export default function PageDevelopers() {
                   defaultValue={filtersByIssueState[0]}
                   options={filtersByIssueState}
                   onChange={handleChangeFilterByState}
+                  styles={{
```
moshmage, line 104: this can easily be done via scss without having to resort to javascript-styles, mainly because we already mutilating the styles of this component

moshmage (changes requested):

styles/styles.scss
```diff
@@ -637,15 +637,15 @@ label {
...
-.color-green {
-  color: $green;
-}
 .color-purple {
   color: $purple;
 }
 .color-red {
   color: $red;
 }
+.color-success {
```
moshmage, line 646: .text-success class already exists

moshmage (approved):

PR #92, opened 2021-09-15
Bugfix/visual bugs 1st setember

clarkjoao:


components/transaction-stats.tsx
```diff
@@ -4,14 +4,18 @@ import {TransactionStatus} from '@interfaces/enums/transaction-status';
 export default function TransactionStats({status = null}: {status: TransactionStatus}) {
   const ColorMap = {
     [TransactionStatus.pending]: `warning`,
-    [TransactionStatus.processing]: `info`,
+    [TransactionStatus.processing]: `blue-info`,
```
moshmage, line 7: info is a bootstrap prefix, if we don't have the correct color for info then we should be editing _the variable of that colour_ and not adding a new info

also, the pending state is repeated

components/transaction-stats.tsx
```diff
@@ -4,14 +4,18 @@ import {TransactionStatus} from '@interfaces/enums/transaction-status';
...
+    [TransactionStatus.pending]: `warning`,
+    [TransactionStatus.failed]: `red`,
     [TransactionStatus.completed]: `success`,
   }
 
   const StatusMap = {
-    [TransactionStatus.completed]: `Complete`,
     [TransactionStatus.pending]: `Pending`,
     [TransactionStatus.processing]: `Processing`,
+    [TransactionStatus.pending]: `Pending`,
```
moshmage, line 16: pending is repeated

styles/styles.scss
```diff
@@ -212,6 +212,21 @@ p {
 .text-blue-dark{
   color: $blue-dark;
 }
+.text-blue-info{
```
moshmage, line 215: override the bootstrap $theme variables instead of adding new classes

styles/styles.scss
```diff
@@ -651,8 +669,13 @@ label {
 .color-red {
   color: $red;
 }
+
+.color-orange {
```
moshmage, line 673: add this to bootstraps utilities of `text-[colorname]`

styles/toasts.scss
```diff
@@ -41,7 +41,8 @@
         justify-content: center;
         margin-right: .15rem;
         &::before {
-          content: "\1F5D9";
+          // content: "\1F5D9";
```
moshmage, line 44: ?!?!?!

moshmage (changes requested):

next-env.d.ts
```diff
@@ -1,3 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
```
moshmage, line 6: ... This link explains nothing about why this file can't be edited and thus this comments adds no real information. no need for the comment to exist

clarkjoao: when I run the project, this line is always added... I will remove

styles/styles.scss
```diff
@@ -212,6 +212,22 @@ p {
...
+}
+
+.text-red {
+  color: $red;
+}
+
+.text-orange {
+  color: $orange;
+}
+.text-success {
```
moshmage, line 227: replace this values on the bootstrap variables

styles/styles.scss
```diff
@@ -212,6 +212,22 @@ p {
...
+.text-info{
+  // !important to overwrite bootstrap css.
+  color: $blue-info !important;
+}
+
+.text-red {
+  color: $red;
+}
+
+.text-orange {
```
moshmage, line 224: add this to text color utilities

styles/styles.scss
```diff
@@ -212,6 +212,22 @@ p {
 .text-blue-dark{
   color: $blue-dark;
 }
+.text-info{
+  // !important to overwrite bootstrap css.
+  color: $blue-info !important;
+}
+
+.text-red {
```
moshmage, line 220: use text-danger on the element

styles/styles.scss
```diff
@@ -212,6 +212,22 @@ p {
 .text-blue-dark{
   color: $blue-dark;
 }
+.text-info{
+  // !important to overwrite bootstrap css.
+  color: $blue-info !important;
```
moshmage, line 217: change the value on the bootstraps variables

moshmage (changes requested):

PR #93, opened 2021-09-16
add: new colors to utilities

clarkjoao:


styles/variables.scss
```diff
@@ -10,29 +10,44 @@ $blue: #4250e4;
...
+        'blue': $blue,
+        'orange': $orange, 
+        'success': $success
+);
+
+$more-colors: (
+        'gray': $gray,
+        'orange': $orange,
+        'purple': $purple,
+        'success': $success,
```
moshmage, line 40: this one doesn't need to be added here, instead it should be overriden on $theme-colors variable

PR #127, opened 2021-09-24
fix: bugs-23-st

clarkjoao:


styles/styles.scss
```diff
@@ -899,6 +899,11 @@ $c_props: ('m': 'margin') !default;
   &-smallest { font-size: 12px; }
   &-small { font-size: 14px; }
 }
+//bg-white-hover from bootstrap, not have --bg-bg-opacity defined.
+.bg-white-hover:hover{
+  background-color: rgba($white, 0.2) !important;
+  color: white !important;
+}
```
clarkjoao, line 906: I rewrote this class because I couldn't change the var '--bg-bg-opacity' in the bootstrap, without this variable, the effect is not applied.
 
Screen Shot 2021-09-24 at 13 44 49

PR #200, opened 2021-11-05
Bugfix/locking ammount updat

clarkjoao:


components/oracles-actions.tsx
```diff
@@ -79,10 +80,16 @@ function OraclesActions(): JSX.Element {
   }
 
   function onSuccess() {
+    setTokenAmount(0);
```
moshmage, line 83: We don't need to change the amount to 0, just the changeBalance update is enough

moshmage (changes requested):

PR #272, opened 2021-12-02
Review pr flow

vhcsilva:


styles/styles.scss
```diff
@@ -427,6 +427,7 @@ p {
     padding: $m;
     margin-top: $m;
     text-align: center;
+    border-radius: $m / 2;
```
moshmage, line 430: use math.div

x-hooks/use-api.tsx
```diff
@@ -270,6 +270,12 @@ export default function useApi() {
       })
   }
 
+  async function createReviewForPR(issueId: string, pullRequestId: string,  githubLogin: string, body:string) {
+    return client.put('/api/pull-request/review', {issueId, pullRequestId, githubLogin, body})
+      .then(response => response)
+      .catch(error => { throw error })
```
moshmage, line 276: This already happens (as in: this catch function is the default one), you don't need to - you can just catch on the other side when you use the function

x-hooks/use-merge-data.ts
```diff
@@ -81,5 +81,28 @@ export default function useMergeData() {
     return pullRequestsWithIssueData
   }
 
-  return {getIssue, getIssues, getPendingFor, getIssuesOfUserPullRequests}
+  async function getMergedDataFromPullRequests(repo, pullRequests) {
+    let mergedPRs = []
```
moshmage, line 85: mergedPRs never changes so it should be a constant

moshmage (approved):

PR #276, opened 2021-12-03
Feature/redesign new proposal

MarcusviniciusLsantos:
- Redesign 
image


components/create-proposal.tsx
```diff
@@ -50,10 +56,81 @@ export default function NewProposal({
...
+     const { id }  = pullRequests.find(
+        (data) => data.githubId === currentGithubId
+      )
+      var currentDistrbuition = {
+        currentPrId: id,
+        prAmounts: participants.map(
+          (items) =>  ((amountTotal * obj[items.githubHandle])/100).toString()
+        )
+      }
+      const equals = (a, b) => JSON.stringify(a) === JSON.stringify(b);
```
moshmage, line 101: This wont work: 

```
{
  a: 1,
  b: 2,
}
```

should be equal to 

```
{
  b: 2,
  a: 1,
}
```

But if we stringify these values, these objects won't match. You'll have to iterate on one and check that all props and values match on the 2nd

components/create-proposal.tsx
```diff
@@ -50,10 +56,81 @@ export default function NewProposal({
...
+        setWarning(' ')
+        setError('')
+        setSuccess('')
+      }else {
+        setError('')
+        setSuccess(' ')
+        setWarning('')
+      } 
+   }
+   if (currentAmount > 0 && currentAmount < 100){
```
moshmage, line 118: these three conditions can be simplified,

```
setError(currentAmount > 0 ? ` ` : ``)
setSuccess(``)
setWarning(``)
```

Also, double-check you're not missing any error messages.

components/input-number.tsx
```diff
@@ -58,7 +70,7 @@ export default function InputNumber({
           {helperText}
         </div>
       )}
-      {error && (
+      {error && error != ' ' && (
```
moshmage, line 73: this is a mighty odd condition. I know what's going on because I just read the code and thus I know what this is. ... a week from now? I'll think this is a bug.

Please, if you're short-circuiting logic use readable names or helping variables.

This means, either introduce a new "shortCircuitError: boolean" variable that's set to true or false when needed and use it here _or_ use a error-code-value to make the logical statement.

moshmage (changes requested):

moshmage (approved):

PR #282, opened 2021-12-09
Feature/adding new route to landing page

MarcusviniciusLsantos:


pages/api/user/number-users.ts
```diff
@@ -0,0 +1,47 @@
+import models from "@db/models";
+import { NextApiRequest, NextApiResponse } from "next";
+import Cors from "cors";
+
+function initMiddleware(middleware) {
```
moshmage, line 5: place this on the `helpers` folder

pages/api/user/number-users.ts
```diff
@@ -0,0 +1,47 @@
...
+          return reject(result);
+        }
+        return resolve(result);
+      });
+    });
+}
+
+const cors = initMiddleware(
+  Cors({
+    methods: ["GET", "POST", "OPTIONS"],
```
moshmage, line 19: `POST` method is not needed here, since used method is `GET`

moshmage (approved):

PR #283, opened 2021-12-10
Feature/modal user missing information

MarcusviniciusLsantos:
- Create a new user warning containing "our account is missing some information, we need you to reconnect your git account"
- Show this to the user when the user connects, has a github account BUT DOES NOT have a accessToken


components/user-missing-information.tsx
```diff
@@ -0,0 +1,57 @@
+import { ApplicationContext } from "@contexts/application";
```
moshmage, line 1: please, use the already existing language functions to translate this modal / page

MarcusviniciusLsantos: done

pages/api/user/[address].ts
```diff
@@ -0,0 +1,31 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import models from "@db/models";
+
+async function remove(req: NextApiRequest, res: NextApiResponse) {
+  const { address } = req.query;
```
Comment (moshmage) [inline pages/api/user/[address].ts:5]: query needs more information provided so it can't be queried by bad actors. We should receive a combo of address+github[name/login/or both]

Comment (moshmage) [reply pages/api/user/[address].ts:5]: - if user has locked ORACLES amount DENY removal

moshmage (changes requested):
Added new information on the Notion issue which will change this PR, please review those changes.

moshmage (approved):

PR #288, opened 2021-12-15
Fix/removing cors logic

MarcusviniciusLsantos:


moshmage (changes requested):
Code has remnants of other work

moshmage (approved):

PR #293, opened 2021-12-21
validate disputable time

vhcsilva:


pages/proposal.tsx
```diff
@@ -43,6 +44,15 @@ interface usersAddresses {
   percentage: number;
 }
 
+export function isProposalDisputable(createdAt: string, disputableTime: number): boolean {
```
moshmage, line 47: export this from the @helpers

pages/developers.tsx
```diff
@@ -67,6 +68,7 @@ export default function PageDevelopers() {
   }
 
   function getIssues() {
+    BeproService.getDisputableTime().then(console.log)
```
moshmage, line 71: left-over from debug

moshmage (changes requested):

PR #385, opened 2022-04-05
Docker Image Build and Deploy Script

hvasconcelos:


scripts/deploy.js
```diff
@@ -0,0 +1,98 @@
+const { Web3Connection, ERC20, NetworkFactory, Network } = require("@taikai/dappkit");
+
+const argv = require('yargs')
+    .alias('e', 'envFile')
+    .describe('e', 'Environment file to load')
+    .default('e', '.env')
+    .alias('n', 'network')
+    .describe('n', 'Ethereum Blockchain')
+    .default('n', 'development')
```
moshmage, line 9: wouldn't it be better to have the user provide a rpc url directly, instead of relying on a hardcoded value that has to be mantaigned ?

scripts/deploy.js
```diff
@@ -0,0 +1,98 @@
...
+    url: `https://ropsten.infura.io/v3/${process.env.DEPLOY_INFURA_KEY}`
+  },
+  seneca: {
+    url: 'http://eth-seneca.taikai.network:8545/',
+  }
+}
+
+async function main() {
+
+  const ownerAddress = process.env.DEPLOY_OWNER_ADDRESS;
```
moshmage, line 44: would be cool to have these (`DEPLOY_OWNER_ADDRESS` and `DEPLOY_PRIVATE_KEY`) as an argument as well, as we would concentrate every option on the entry command ^^

ganache/deploy.js
```diff
@@ -3,7 +3,7 @@ const {
   ERC20,
   toSmartContractDecimals,
   Network
-} = require("bepro-js");
+} = require("@taikai/dappkit");
```
moshmage, line 6: I'd say this script is moot since we're introducing the same script on `scripts/deploy.js` ? can we remove this?

@clarkjoao @vhcsilva

package.json
```diff
@@ -16,13 +16,18 @@
     "start": "next start",
     "make-seo": "node make-seo.js",
     "schedules": "node schedules.js",
-    "lint": "npx eslint . --fix --ext .ts,.tsx"
+    "lint": "npx eslint . --fix --ext .ts,.tsx",
+    "deploy:local": "node ./scripts/deploy -n development",
```
moshmage, line 20: and if we can have the `script/deploy.js` receive the rpcUrl and all the needed arguments then we don't need to have all these scripts (nor keep adding more) because we can simply run a command :D

moshmage (changes requested):

package.json
hvasconcelos: it is hardhat like script. I used the same approach than other projects like uniswap . you dont have to remember the RPC . Just type the command and the rest is magic.

scripts/deploy.js
hvasconcelos: it think it would be great to have to move the networks variable to a file with a list of networks.

scripts/deploy.js
hvasconcelos: yes i can do it.

ganache/deploy.js
vhcsilva: Yes for me

scripts/deploy.js
hvasconcelos: Supports passed url and statically defined networks

ganache/deploy.js
moshmage: @hvasconcelos remove this file please :)

scripts/deploy.js
```diff
@@ -1,15 +1,56 @@
...
+    .option('a', {
+      alias: "ownerAddress",
+      describe: "Owner Eth Address",
+      type: 'string',
+      nargs: 1,
+    })  
     .alias('h', 'help')
     .help('help')
-    .usage('Usage: $0 [options] <file>')
+    .usage(usage)
```
moshmage, line 53: I'm not sure, but I would think that if you don't provide this `yarn` will provide the `help` thing -- that way we don't have to write one by hand ;)

moshmage (approved):

PR #394, opened 2022-04-20
Distribute bepro on Deploy scripts

hvasconcelos:


scripts/staging-accounts.js
```diff
@@ -0,0 +1,26 @@
+module.exports = {
```
moshmage, line 1: we don't need to export the private keys (while they _are_ from a private chain) since we only need the addresses themselves, we can simply export an array of addresses

moshmage (approved):

PR #406, opened 2022-05-03
Vhc 279 using GitHub graphql

vhcsilva:


interfaces/authentication.ts
```diff
@@ -8,9 +8,16 @@ export interface IUser {
   accessToken?: string;
 }
 
+export interface IRepository {
```
moshmage, line 11: use "RepositoryInfo" avoid `I` for interfaces

interfaces/authentication.ts
```diff
@@ -8,9 +8,16 @@ export interface IUser {
...
+  name: string;
+  nameWithOwner: string;
+  isFork: boolean;
+}
+
 export interface IWallet {
```
moshmage, line 17: Same here, avoid `I` for interfaces

package.json
```diff
@@ -28,6 +28,7 @@
     "ganache:deploy": "node ganache/deploy.js"
   },
   "dependencies": {
+    "@octokit/graphql": "^4.8.0",
```
moshmage, line 31: was the other ockokit we were using removed?

x-hooks/use-octokit-graph.tsx
```diff
@@ -0,0 +1,293 @@
...
+    const { additions, deletions } = response["repository"]["pullRequest"];
+
+    return additions + deletions;
+  }
+
+  async function getIssueOrPullRequestComments(repositoryPath:  string, id: number) {
+    const { owner, repo } = getOwnerRepoFrom(repositoryPath);
+
+    const response = await getAllPages(`
+        query Comments($repo: String!, $owner: String!, $id: Int!, $cursor: String) {
```
moshmage, line 123: put these queries into a `graph-ql-queries` folder, and then import them and use them

moshmage (changes requested):

package.json
vhcsilva: @moshmage no, we are using it on BE

moshmage (approved):

helpers/object.ts
```diff
@@ -0,0 +1,9 @@
+export const getPropertyRecursively = (property, data) => {
```
moshmage, line 1: `getPropertyRecursively` has to explicitly take in a type that and return that same type,

`export function getProp<T = any>(prop: string, data = {}): T[keyof T]`

If this isn't possible, then _all variables that use this function must explicitly say what their type is_

pages/api/past-events/move-to-open/index.ts
```diff
@@ -53,19 +56,41 @@ async function post(req: NextApiRequest, res: NextApiResponse) {
...
+        if (currentRepo !== `${owner}/${repo}`) {
+          currentRepo = `${owner}/${repo}`;
+
+          const repositoryDetails = await githubAPI(RepositoryQueries.Details, {
+            repo,
+            owner
+          });
+
+          labelId = 
+            repositoryDetails["repository"]["labels"]["nodes"].find(label => label.name.toLowerCase() === "draft")?.id;
```
moshmage, line 82: why are we using `[]` notation? can't we simply use `.notation` (dot notation?)

pages/api/past-events/redeem-issue/index.ts
```diff
@@ -58,13 +61,21 @@ async function post(req: NextApiRequest, res: NextApiResponse) {
...
+
+        const issueDetails = await githubAPI(IssueQueries.Details, {
           repo,
-          issue_number: +issue.githubId,
-          state: "closed"
+          owner,
+          issueId: +issue.githubId
         });
+      
+        const issueGithubId = issueDetails["repository"]["issue"]["id"];
```
moshmage, line 73: Same here, use dot notation

types/octokit.d.ts
```diff
@@ -0,0 +1,5 @@
+export type GraphQlQueryResponseData = {
```
moshmage, line 1: I'm betting that GraphQL provides their own types, look into it

x-hooks/use-merge-data.ts
```diff
@@ -17,7 +15,7 @@ const OctoData = {};
 export default function useMergeData() {
   const { loadRepos, repoList } = useRepos();
   const db = useApi();
-  const octokit = useOctokit();
+  const octokit = null;
```
moshmage, line 18: ?

moshmage (changes requested):

helpers/object.ts
```diff
@@ -0,0 +1,9 @@
+export function getPropertyRecursively<T> (property: string, data = {}): T[keyof T] {
+  if (data[property]) return data[property];
+
+  let found = {};
+
+  for(const key in data) found = getPropertyRecursively(property, data[key]);
```
moshmage, line 6: We need to make this for short-circuit itself, or we are going to load forever,

```javascript
if (data[property]) return data[property];
let found = null;
const keys = Object.keys(data);
for (let x = 0; x <= keys.length - 1 && !found; x++) {
  found = getPropertyRecursively(property, keys[x]);
}
return found;
```

pages/api/past-events/move-to-open/index.ts
```diff
@@ -94,8 +121,8 @@ async function post(req: NextApiRequest, res: NextApiResponse) {
 export default async function MoveToOpen(req: NextApiRequest,
                                          res: NextApiResponse) {
   switch (req.method.toLowerCase()) {
-  case "post":
-    await post(req, res);
+  case "get":
```
moshmage, line 124: doesn't this endpoint change the state? why did we change it to a GET ?

moshmage (approved):

PR #407, opened 2022-05-03
Vhc 298 fails to create network

vhcsilva:
- using defaults for treasury address, cancel fee and close fee until the design is ready
- added a new environment variable for the nft token to be used as default


services/bepro-service.ts
```diff
@@ -148,11 +148,14 @@ class BeproFacet {
   }
 
   async createNetwork(networkToken: string = publicRuntimeConfig.contract.settler, 
-                      nftToken = '0x23891F76C9A3Ad7c523F4A373881bD7D282B5D5E', 
-                      nftUri = '//') {
+                      nftToken: string = publicRuntimeConfig.contract.nft, 
+                      nftUri = '//',
```
moshmage, line 152: [dev-268](https://taikai.atlassian.net/jira/software/c/projects/DEV/boards/44?modal=detail&selectedIssue=DEV-268) -- we need this nftUri to point to somewhere; Im thinking this should be configurable, with our default value set in. This can be made on other PR, just pointing it out :)

moshmage (approved):

PR #411, opened 2022-05-09
Proposal Merge - Show distribution percentages

MarcusviniciusLsantos:


helpers/calculateDistributedAmounts.ts
```diff
@@ -0,0 +1,25 @@
...
+import { BeproService } from "services/bepro-service";
+
+
+export default async function calculateDistributedAmounts(bountyAmount: number, proposalPercents: number[]) {
+  const treasury = await BeproService.network.treasuryInfo();
+  
+  const treasuryAmount =
+    treasury["0"] === Defaults.nativeZeroAddress
+      ? 0
+      : (bountyAmount / 100) * (treasury["1"] / Defaults.TenK);
```
moshmage, line 12: use `handlePercentage` (as well as on the following lines)

components/proposal-merge.tsx
```diff
@@ -0,0 +1,152 @@
...
+  proposal: ProposalExtended,
+  mergeBounty: () => void;
+  canMerge: boolean
+}
+
+export default function ProposalMerge({
+  amountTotal,
+  tokenSymbol,
+  proposal,
+  mergeBounty,
```
moshmage, line 38: switch naming to `onClickMerge` so we can understand that this is both a function _and_ a "event handler"

moshmage (approved):

moshmage:
@MarcusviniciusLsantos I pressed the wrong button, it should have been "request changes" instead of "approve", my bad. please, checkout the comments and checkout why the build is breaking

PR #413, opened 2022-05-10
Vhc 297 list organizations repositories

vhcsilva:


contexts/authentication.tsx
```diff
@@ -55,14 +54,12 @@ export const AuthenticationProvider = ({ children }) => {
 
       await BeproService.login();
 
-      const [isCouncil, isApprovedSettlerToken] = await Promise.all([
-        BeproService.isCouncil(),
-        BeproService.isApprovedSettlerToken()
+      const [isCouncil] = await Promise.all([
```
moshmage, line 57: no need to Promise.all when you only have one promise :)

pages/new-network.tsx
```diff
@@ -127,7 +127,7 @@ export default function NewNetwork() {
 
     await BeproService.startNetworkFactory();
 
-    BeproService.createNetwork()
+    BeproService.createNetwork(publicRuntimeConfig?.contract?.settler, steps.tokens.nftToken)
```
moshmage, line 130: Shouldn't the user be free to choose the settler token? :o

components/custom-network/network-information-step.tsx
```diff
@@ -41,7 +41,7 @@ export default function NetworkInformationStep({
     const name = e.target.value;
     const exists =
       name.trim() !== ""
-        ? name.toLowerCase().includes("bepro")
+        ? (name.toLowerCase().includes("bepro") || name.toLowerCase().includes("taikai"))
```
moshmage, line 44:
```
['taikai', 'bepro'].includes(name.toLowerCase())
```

components/custom-network/token-configuration.tsx
```diff
@@ -55,29 +55,42 @@ export default function TokenConfiguration({
...
     changedDataHandler("tokens", { label: "nftToken", value: address });
   }
 
-  async function validateAddress() {
+  async function validateNFTAddress() {
     if (data.nftToken.trim() === "") return changedDataHandler("tokens", { label: "validated", value: false });
 
-    try {
-      const token = new BountyToken(BeproService.bepro, data.nftToken);
+    const token = new BountyToken(BeproService.bepro, data.nftToken);
```
moshmage, line 72: So, if data.nftToken _is not_ a BountyToken contract, this will throw an error that you're not catching. We need to catch that error and tell it to the user so the user understands _what kind of contract it needs_ because they could copy-paste a NFT address and think it will work

components/custom-network/token-configuration.tsx
```diff
@@ -55,29 +55,42 @@ export default function TokenConfiguration({
...
+    changedDataHandler("tokens", { label: "networkToken", value: token.address });
+  }
+
   function setDeployedAddress(address) {
-    changedDataHandler("tokens", { label: "validated", value: true });
     changedDataHandler("tokens", { label: "nftToken", value: address });
   }
 
-  async function validateAddress() {
+  async function validateNFTAddress() {
```
moshmage, line 69: You can define this function inside the useEffect since I don't think its used anywhere else.

services/bepro-service.ts
```diff
@@ -191,6 +191,12 @@ class BeproFacet {
     return network.getOraclesResume(this.address);
   }
 
+  async getSettlerTokenAllowance(networkAddress: string = publicRuntimeConfig?.contract?.address): Promise<number> {
+    const network = await this.getNetworkObj(networkAddress);
+    
+    return this.getAllowance(network.settlerToken.contractAddress, this.address, networkAddress);
```
moshmage, line 197: _please_, for everyones' sanity, don't call a function that was not declared yet:

```
function B() {
  A(); // I don't know what A does, as it has not been declared on the code yet
}

function C() {
  A() // Same
}

function A() {} // I needed to get to the end of the file so I can read this declaration and now scroll back up to make sense of the logic inside the functions
```

but if you declare `A()` first, then everything falls into place :)

components/custom-network/network-information-step.tsx
vhcsilva: for this case i need to know if the words "bepro" or "taikai" are in the name, the way you suggested we are going to verify if the name matches these words

components/custom-network/network-information-step.tsx
```diff
@@ -38,12 +39,11 @@ export default function NetworkInformationStep({
...
-    const exists =
-      name.trim() !== ""
-        ? name.toLowerCase().includes("bepro")
-          ? false
-          : !(await networkExists(name))
-        : undefined;
+
+    let exists = undefined;
+
+    if (name.trim() !== "")
```
moshmage, line 45: LOL it wasn't simple to start with, but you went further and beyond xD

moshmage (approved):

PR #422, opened 2022-05-18
creating new service and context

vhcsilva:


services/dao-service.ts
```diff
@@ -0,0 +1,155 @@
...
+  private _network: Network_v2;
+
+  get web3Connection() { return this._web3Connection; }
+  get network() { return this._network; }
+
+  constructor() {
+    if (!publicRuntimeConfig?.web3ProviderConnection) 
+      throw new Error("Missing web3ProviderConnection in next.config.js");
+
+    this._web3Connection = new Web3Connection({
```
moshmage, line 23: you can have this prop as a `readonly` and assign value to it without a constructor, 

```
readonly web3Connection = new Web3Connection({ /* ... */});

constructor() {
    if (!publicRuntimeConfig?.web3ProviderConnection) 
      throw new Error("Missing web3ProviderConnection in next.config.js");
}
```

typescript will essentially do what you're doing, but provide "web3connection" as writable. Since we're using Typescript _we_ won't be able to change that value through code or lint will scream bloody murder -- and that's essentially what we want; we don't need _an actual readonly value_ because no one, but us, is consuming this service ;)

ps, [proof](https://www.typescriptlang.org/play?#code/DYUwLgBAZghglsCBeaNgGcQG4BQODGwM66EAghAN44S0QBOIMAJgPYB2wAnuchAIy46EfB3Rh6AV3xhW9ABQBKKjWF04UCPNgJFqtXTAALeqwDuEdiAsBReqYUByHaGaPFQugF8cPnAHp-CDMERAA5AHkAFVQEAjFIGD4rCzIlXFF2dFZQADpgVgBzeRhcsmVAgTwXPglJbBwJHmphTPEIACNk63J03xEYMHwjLRBlFro2nJB8ovlHAGEyAFUAcQAJGIAuABF3T1opvILisawISvxB4YhZCABbGABrEAh0SUYAQl8gA)

services/dao-service.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  async getAddress(): Promise<string> {
+    return this.web3Connection.getAddress();
+  }
+
+  async getChainId(): Promise<number> {
+    return this.web3Connection.web3.eth.getChainId();
+  }
+
+  async getNetwork(networkAddress: string = undefined): Promise<Network_v2> {
```
moshmage, line 81: This function can't be called this many times; If we're rewriting this, we need to make sure that when we `new` this class, we have to make sure that a network is already loaded.

components/page-actions.tsx
```diff
@@ -171,7 +171,7 @@ export default function PageActions({
   };
 
   const renderUpdateAmount = () => {
-    if (isIssueinDraft && isBountyOwner())
+    if (isIssueinDraft && isBountyOwner() && !finalized)
```
moshmage, line 174: how can a bounty be in draft _but be finalized_ ?

components/proposal-merge.tsx
```diff
@@ -54,20 +58,32 @@ export default function ProposalMerge({
...
-    getDistributedAmounts();
-  }, [proposal]);
+  const { activeNetwork } = useNetwork();
 
+  
   async function getDistributedAmounts() {
     if (!proposal?.details) return;
+    
+    const distributions = calculateDistributedAmounts(activeNetwork?.treasury,
+                                                      activeNetwork?.mergeCreatorFeeShare,
```
moshmage, line 68: lint will kill you here; You can escape it by breaking line after the `=` sign on 67 and then indent each line with 2 spaces:

```ts
const distributions =
  calculateDistributedAmounts(activeNetwork?.treasury,
    activeNetwork?.mergeCreatorFeeShare, // use 4 spaces here
    /* ... */);
```

contexts/network.tsx
```diff
@@ -67,42 +66,59 @@ export const NetworkProvider: React.FC = function ({ children }) {
...
+  }, [query?.network]);
+
+  useEffect(() => {
+    if (activeNetwork) updateNetworkParameters();
+  }, [DAOService, activeNetwork]);
 
   useEffect(() => {
-    if (query?.network) updateNetworkParameters();
-  }, [beproServiceStarted, query?.network]);
+    if(activeNetwork?.networkAddress) changeNetwork(handleNetworkAddress(activeNetwork));
```
moshmage, line 113: should we attest that activeNetwork is not the activeNetwork ? (so we dont change into the same network, triggering the useEffect again?)

services/dao-service.ts
```diff
@@ -0,0 +1,483 @@
...
+    return this.factory.createNetwork(networkToken, nftToken, nftUri, treasuryAddress, cancelFee, closeFee);
+  }
+
+  async getNetworkAdressByCreator(address: string): Promise<string> {
+    if (!this.factory) await this.loadFactory();
+
+    return this.factory.networkOfAddress(address);
+  }
+
+  // TODO: use this method when the https://github.com/taikai/dappkit/pull/25 is merged
```
moshmage, line 455: this was already merged :)

moshmage (approved):

PR #423, opened 2022-05-19
Bugfix/try upload logo

clarkjoao:


moshmage (approved):

pages/api/network/index.ts
```diff
@@ -207,12 +216,21 @@ async function put(req: NextApiRequest, res: NextApiResponse) {
     }
 
     // Uploading logos to IPFS
-    const fullLogoHash = fullLogo
-      ? (await IpfsStorage.add(fullLogo, true, undefined, "svg")).hash
-      : undefined;
-    const logoIconHash = logoIcon
-      ? (await IpfsStorage.add(logoIcon, true, undefined, "svg")).hash
-      : undefined;
+    let fullLogoHash = null
```
moshmage, line 219: is this repeated code ?

clarkjoao: yep, are different methods.

pages/api/network/index.ts
```diff
@@ -280,8 +263,20 @@ async function put(req: NextApiRequest, res: NextApiResponse) {
 
     if (!isAdminOverriding) network.colors = JSON.parse(colors);
 
-    if (logoIconHash) network.logoIcon = logoIconHash;
-    if (fullLogoHash) network.fullLogo = fullLogoHash;
+    if (fullLogo || logoIcon) {
```
moshmage, line 266: This can be made into a function and be reused on line 79

pages/[network]/account/payments.tsx
```diff
@@ -66,7 +66,7 @@ export default function Payments() {
   const {getPayments} = useApi()
   const {wallet} = useAuthentication()
   const [payments, setPayments] = useState<Payment[]>([]);
-  const [hasMore, setHasMore] = useState(false);
+  const [hasMore] = useState(false);
```
Comment (moshmage) [inline pages/[network]/account/payments.tsx:69]: do we ever change "hasMore"? If not -- is it really needed?

scripts/deploy-v2.js
```diff
@@ -133,7 +133,7 @@ async function main() {
    const beproToken = new ERC20(web3Connection, beproAddress);
    await beproToken.start();
 
-   if(argv.network !== 'custom')
+   if(argv.network !== 'custom' || argv.network !== 'local')
```
moshmage, line 136: why do we want to _not_ transfer when we are on a local network ?

moshmage (commented):
Some doubts, but overall lgtm

scripts/deploy-v2.js
clarkjoao: because these are the test wallets on the test network, on the local network these wallets do not have access to the ganache, so it makes no sense to transfer to them.

pages/[network]/account/payments.tsx
Comment (clarkjoao) [reply pages/[network]/account/payments.tsx:69]: we still don't use hasMore, but we will use it to paginate the payments, I removed the 'set' temporarily to correct the tsconfig warnings.

moshmage (approved):

PR #429, opened 2022-06-07
Vhc 325 use network registry

vhcsilva:


components/custom-network/treasury-step.tsx
```diff
@@ -0,0 +1,113 @@
...
+  function handleCloseFeeChange(e) {
+    setCloseFee(e.target.value);
+  }
+
+  function handleCloseFeeBlur() {
+    fields.closeFee.setter(address);
+  }
+  
+  return(
+    <Step
```
moshmage, line 49: treasury is optional, remember that

contexts/network-settings.tsx
```diff
@@ -0,0 +1,434 @@
...
+        validated: false, 
+        name: { 
+          value, 
+          validated: undefined 
+        } 
+      })),
+      validator: async (value: string): Promise<boolean | undefined> => {
+        let validated = undefined;
+  
+        if (value.trim() !== "")
```
moshmage, line 80: I've noticed we're allowing _spaces_ on the network name (There's a network literally called `"network name"` -- we need to replace the spaced into dashes, after trimming:

```js
value = value.trim().replace(/\s+/g,'-');
```
```

File: styles/components/amount-with-preview.scss [hunk 3]
```diff
@@ -0,0 +1,30 @@
...
+    svg {
+      path {
+        fill: var(--bs-danger);
+      }
+    }
+  }
+  &.text-purple {
+    svg {
+      path {
+        fill: $purple;
```
moshmage, line 26: no --bs-purple? :(

contexts/network-settings.tsx
vhcsilva: @moshmage do we need to replace just because of the URL or really the name can't have spaces? if it's just because of the URL, I can make the replacement only at the time of assembling the url

components/custom-network/treasury-step.tsx
vhcsilva: yes, it's working as optional

contexts/network-settings.tsx
moshmage: @vhcsilva both -- if we allow spaces but replace the dash, there could be something like "new network" _and_ "new-network" that would be different but be shown the same way and then would create a conflict when we try to query it

moshmage (approved):

moshmage (approved):

PR #430, opened 2022-06-13
Remock github response on web-network-e2e

MarcusviniciusLsantos:


services/dao-service.ts
```diff
@@ -473,4 +473,16 @@ export default class DAO {
   isAddress(address: string): Promise<boolean> {
     return this.web3Connection.utils.isAddress(address);
   }
+
+  getTimeChain(): Promise<number> { 
+    return this.web3Connection.Web3.eth.getBlock(`latest`).then(block => block.timestamp*1000);
+  }
+
+  async isBountyInDraftChain(creationDateIssue: number): Promise<boolean> {
```
moshmage, line 481: I would wager this function exists on the NetworkV2

pages/api/_middleware.ts
```diff
@@ -1,10 +1,13 @@
 import type { NextApiRequest } from 'next'
 import { getToken } from 'next-auth/jwt'
+import getConfig from 'next/config';
 import { NextResponse } from 'next/server'
 
 import { error, info } from 'helpers/api/handle-log'
 
+const { serverRuntimeConfig } = getConfig();
```
moshmage, line 8: and then you'll be able to deconstruct it here without any `?` null coalescent needed

```ts
const {serverRuntimeConfig: { e2e } = getConfig();
```

pages/api/_middleware.ts
```diff
@@ -19,7 +22,7 @@ export async function middleware(req: NextApiRequest) {
     info({method, ip, ua, ...page, pathname, search, body});
 
 
-  if(method !== 'GET' && !isInWhiteList){
+  if(method !== 'GET' && !isInWhiteList || serverRuntimeConfig?.e2e?.testnet !== true){
```
moshmage, line 25: Note: this condition will always be true in production and thus it won't respect the previous condition. What exactly is it that you want to achieve with this condition ?

moshmage (approved):

PR #432, opened 2022-06-18
Dev 416 allow pull request to be open even if bounty is ready

vhcsilva:


helpers/api/proposal/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { Network_v2 } from "@taikai/dappkit";
+import { Bounty } from "@taikai/dappkit";
+
+/**
+ * @returns * -1: Error
```
moshmage, line 5: -1 should `throw new Error()` and be typedoced as such,

```ts
/**
 * @returns <number> 0: not ready, 1: invalid, 2: valid, 3: no proposals
 * */

moshmage (approved):

PR #437, opened 2022-06-20
DEV-276 create amount of council members

clarkjoao:


moshmage (approved):

helpers/api/oracles/update-amount.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+  })
+
+  if (!network) return console.warn("Network not found in the database", network?.contractAddress);
+
+
+  try {
+    for (const event of events) {
+      const { newLockedTotal, actor } = event.returnValues;
+      const councilAmount = +fromSmartContractDecimals(newLockedTotal) > (await network?.councilAmount());
```
moshmage, line 19: There's too many operations going inside this for loop, and even more loops; We don't need this many "inner loops":

```ts

const councilAmount = await network.councilAmount;
const new_members = [];
const existing_members = []; // _network.councilMembers
const remove_members = [];

for (const event of events) {
  const {newLockedTotal, actor} = event.returnValues;
  if (newLockedTotal >= councilAmount && !existing_members.includes(actor))
    new_members.push(actor)
  else if (newLockedTotal < councilAmount && existing_members.includes(actor))
    remove_members.push(actor);
}

existing_members
  .map((address, index) => remove_members.includes(actor) && index)
  .forEach(index => delete _network.councilMembers[index]);

new_members.forEach(address => _network.coundilMembers.push(address))

_network.save();
```

We do the same loops, only we never do them "inside" another loop, so we are always O1 :D

moshmage (changes requested):

helpers/api/oracles/update-amount.ts
```diff
@@ -0,0 +1,37 @@
+import { Network_v2, fromSmartContractDecimals } from "@taikai/dappkit";
+
+import models from "db/models";
+export default async function updateAmount(events, network: Network_v2) {
```
moshmage, line 4: rename to `updateOraclesAmount` because while _it is_ on the oracles folder, while reading the code you might forget that

moshmage (changes requested):

moshmage (approved):

PR #446, opened 2022-06-27
Dev 369 if user cant access repo the page should mirror that show a prohibiting modal

clarkjoao:


helpers/object.ts
```diff
@@ -1,13 +1,17 @@
...
+  if (data?.[property]) return data[property];
 
   let found = null;
 
-  const keys = Object.keys(data);
+  if(data){
+    const keys = Object?.keys(data)
+  
+    if(keys?.length > 0) 
+      for (let x = 0; x < keys.length && !found; x++) {
```
moshmage, line 10: you can simply add `?.` to the for-loop since it will evaluate to `undefined` and `undefined` is casted to `false` and then to `0` because ... javascript,

```js
var a = undefined;
for (let i = 0; i < a; i++) {
  console.log('acted', i+1, 'times')
} 
```

PR #450, opened 2022-06-28
Dev 516 decouple GitHub from connect button and enable users to use only a wallet account

vhcsilva:


components/proposal-merge.tsx
```diff
@@ -80,7 +80,7 @@ export default function ProposalMerge({
   async function  getCoinInfo() { 
     await getCoinInfoByContract(activeNetwork?.networkToken?.address).then((tokenInfo) => {
       setCoinInfo(tokenInfo)
-    }).catch(err => console.error('CoinInfo', err))
+    }).catch(() => ({}));
```
moshmage, line 83: log the error out, just so we know something happened, so we can debug it in the future if something eventually breaks here

x-hooks/use-octokit.tsx
```diff
@@ -17,7 +17,7 @@ export default function useOctokit() {
   }
 
   function makeOctokitRequest(query: string, params): Promise<GraphQlResponse> {
-    return api.post("/graphql", { query, params }).then(({ data }) => data).catch(error => error.data);
+    return api.post("/graphql", { query, params }).then(({ data }) => data)
```
moshmage, line 20: leave the catch statement, and log the error, your future self might thank you

pages/api/graphql/index.ts
```diff
@@ -1,36 +1,37 @@
...
+      auth: access_token || botToken
     });
 
     const result = await octokit.graphql(query, params);
 
     return res.status(200).json(result);
   } catch(error) {
-    error("", { req, error });
-    return res.status(500).json(error);
+    LogError("", { req, error, access_token, login, name });
```
moshmage, line 24: add `GraphQL Proxy` as a message so we can _eventually_ find it on the elastic search

moshmage (approved):
lgtm, some catches were erased that could still be alive but its not a blocker.

PR #457, opened 2022-07-01
Dev 486 new create bounty modal

MarcusviniciusLsantos:


components/create-bounty-modal.tsx
```diff
@@ -38,21 +46,21 @@ import { IFilesProps } from "./drag-and-drop";
...
-   cid: string | boolean,  
-   repoPath: string,
-   transactional: string,
-   branch: string
-   githubUser: string,
-   tokenAmount: number,
-   rewardToken?: string,
-   rewardAmount?: number,
-   fundingAmount?: number,
+  title: string;
```
moshmage, line 54: Can't we simply use `import {Bounty} from '@taikai/dappkit'` and use `Bounty` interface instead of having to rewrite it ?

pages/[network]/developers.tsx
```diff
@@ -88,7 +88,7 @@ export default function PageDevelopers() {
 export const getServerSideProps: GetServerSideProps = async ({ locale }) => {
   return {
     props: {
-      ...(await serverSideTranslations(locale, ["common", "bounty"]))
+      ...(await serverSideTranslations(locale, ["common", "bounty", "bounty", "connect-wallet-button"]))
```
Comment (clarkjoao) [inline pages/[network]/developers.tsx:91]: why double bounty?

components/create-bounty-token-amount.tsx
```diff
@@ -0,0 +1,134 @@
...
+            label={" "}
+            className="mt-3"
+            symbol={"EUR"}
+            classSymbol="text-white-30 mt-3"
+            disabled={true}
+            value={
+              getCurrentCoin()?.tokenInfo
+                ? handleTokenToEurConversion(Number(issueAmount.value),
+                                             getCurrentCoin()?.tokenInfo?.prices["eur"])
+                : "-"
```
moshmage, line 117: this value makes no sense. Either remove it (and the placeholder takes hold) or put it at 0

moshmage (approved):

components/create-bounty-token-amount.tsx
MarcusviniciusLsantos: This condition is used when the coingecko api does not return information. if you put 0, it won't give the idea that this coin is worth 0 euros, right?

moshmage (approved):

PR #459, opened 2022-07-04
Dev 464 new user menu

vhcsilva:


components/nav-avatar.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+import { useAuthentication } from "contexts/authentication";
+
+import { truncateAddress } from "helpers/truncate-address";
+
+import useNetworkTheme from "x-hooks/use-network";
+
+const Link = (label, href) => ({ label, href });
+
+const ProfileInternalLink = ({ label, href }) =>
```
moshmage, line 19: these templates should be inside the NavAvatar function as a closure, since they are not exported nor used anywhere else - no need for them to be outside of the scope that uses them

contexts/authentication.tsx
```diff
@@ -125,12 +133,16 @@ export const AuthenticationProvider = ({ children }) => {
...
     if (!DAOService) return;
     
-    window?.ethereum?.on("accountsChanged", () => {
-      DAOService.connect()
-        .then(connected => {
-          if (connected) updateWalletAddress();
-        })
-        .catch(error => console.log("Failed to change account", error));
+    window?.ethereum?.on("accountsChanged", (accounts) => {
+      if (window.connectedAddress)
```
moshmage, line 137: `if !DAOService.web3Connection.connected` is what you're looking for.

Don't forget that accounts CAN BE EMPTY (when user disconnects from metamask)

interfaces/window.ts
```diff
@@ -3,5 +3,6 @@ import DAO from "services/dao-service";
 declare global {
   interface Window {
     DAOService: DAO;
+    connectedAddress: string;
```
moshmage, line 6: no need, you can query this off of `DAOService.web3Connection`

moshmage (changes requested):

contexts/authentication.tsx
```diff
@@ -125,14 +130,17 @@ export const AuthenticationProvider = ({ children }) => {
...
+    if (wallet?.address)
+      window?.ethereum?.on("accountsChanged", () => {
+        DAOService.connect()
+          .then(connected => {
+            if (connected) updateWalletAddress();
+          })
+          .catch(error => console.log("Failed to change account", error));
+      });
+    else
+      window?.ethereum?.removeAllListeners("accountsChanged");
```
moshmage, line 142: ?!?!? what.. what if you connect the wallet again? xD

vhcsilva: @moshmage it will enter the if instead of else

PR #465, opened 2022-07-07
Dev 470 new account profile page

vhcsilva:


components/profile/wallet-balance.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const [tokens, setTokens] = useState<TokenBalance[]>([]);
+  const [totalEuro, setTotalEuro] = useState(0);
+  const [hasNoConvertedToken, setHasNoConvertedToken] = useState(false);
+
+  const { activeNetwork } = useNetwork();
+  const { wallet } = useAuthentication();
+  const { service: DAOService } = useDAO();
+
+  const FlexRow = ({ children, className = "" }) => <div className={`d-flex flex-row ${className}`}>{children}</div>;
```
moshmage, line 31: Would be interesting to have this (and Column) components in a component that could be used by everything else :)

moshmage (approved):

components/profile/wallet-balance.tsx
vhcsilva: Yes, it is. I already changed it in the branch I'm working now.

PR #470, opened 2022-07-13
DEV-356 elastic logs

clarkjoao:


middleware/withCors.ts
```diff
@@ -18,8 +21,16 @@ function runMiddleware(req, res, fn) {
...
+function runLogger(req) {
+  const {method, url, ip, ua, body} = {...req} as any; // eslint-disable-line
+  const pathname = url.split('/api')[1];
+  if (!ignorePaths.some(k => pathname.includes(k)))
+    info({method, ip, ua, pathname, body});
+}
+
 const withCors = (handler) => {
   return async (req, res) => {
+    runLogger(req);
```
moshmage, line 33: by using the logger here, we fail to catch errors from the crash of the normal api, by runing it on the middleware we gain the `catch` ability - along with any `throw Error()` being correctly parsed :)

moshmage (changes requested):

helpers/api/handle-log.ts
```diff
@@ -1,3 +1,8 @@
+
+import { Client } from "@elastic/elasticsearch";
+import getConfig from "next/config";
+const { serverRuntimeConfig } = getConfig();
+
 const Levels = {log: 'LOG', info: 'INFO', error: 'ERROR'};
 
 const output = (level: string, message: string|any, rest) => { // eslint-disable-line
```
moshmage, line 8: please, change `rest` to `...rest`

middleware/withCors.ts
moshmage: @clarkjoao I just noticed that all that is needed is on line 38 we need to log the error that was sent via `error(e)`

middleware/withCors.ts
```diff
@@ -18,8 +21,16 @@ function runMiddleware(req, res, fn) {
   })
 }
 
+function runLogger(req) {
+  const {method, url, ip, ua, body} = {...req} as any; // eslint-disable-line
+  const pathname = url.split('/api')[1];
+  if (!ignorePaths.some(k => pathname.includes(k)))
```
moshmage, line 27: you can `url.search(`/health/`) > -1` and we simplify this :)

middleware/withCors.ts
```diff
@@ -18,8 +21,16 @@ function runMiddleware(req, res, fn) {
   })
 }
 
+function runLogger(req) {
+  const {method, url, ip, ua, body} = {...req} as any; // eslint-disable-line
```
moshmage, line 25: you're missing `page = {}`

PR #475, opened 2022-07-15
Dev 469 other profile pages

vhcsilva:


components/custom-network/treasury-address-field.tsx
```diff
@@ -0,0 +1,33 @@
+import { useTranslation } from "next-i18next";
+
+export default function TreasuryAddressField({
+  value,
+  onChange,
+  validated = undefined
+}) {
+  const { t } = useTranslation("custom-netwokr");
```
moshmage, line 8: typo, "netwokr"

services/dao-service.ts
```diff
@@ -97,6 +98,13 @@ export default class DAO {
     try {
       await this.web3Connection.start();
 
+      console.table({
```
moshmage, line 101: yay! :D

services/dao-service.ts
```diff
@@ -49,6 +49,7 @@ export default class DAO {
       await network.loadContract();
 
       if (!skipAssignment) this._network = network;
+      else this._networkAux = network;
```
moshmage, line 52: Like we spoke, it's easier to have the component itself initialize a `new DAO()` and then `loadNetwork('0xCustomNetworkAddress')` than having an auxiliary network _that might_ get confused on the if/else and return the wrong network on the `getNetwork()` getter.

We also ensure that all the methods are called on the network that we want to edit without adding any more complexity :)

moshmage (changes requested):

components/oracle.tsx
```diff
@@ -37,36 +39,52 @@ export default function Oracle({ children }) {
...
+            currency: "BEPRO"
+          },
+          {
+            value: totalUsers,
+            label: t("common:heroes.protocol-members")
+          }
+        ]);
+      });
+    } catch (error) {
+      console.log("Failed to load oracle data", error, activeNetwork);
```
moshmage, line 80: This error should be _seen_ as well -- add a toastError here so we are warned without having to look into the console log (which will also help with understanding when something went wrong)

contexts/network-settings.tsx
```diff
@@ -191,78 +194,97 @@ export const NetworkSettingsProvider = ({ children }) => {
...
-    setSettings(previous => ({
-      ...previous,
-      theme: {
-        ...previous.theme,
-        colors: isCreating && DefaultTheme() || network?.colors,
-      },
-    }));
+          if (isCreating) repositories.push(...filtered);
+          else {
+            let { rows: networkRepositories } = await searchRepositories({ networkName: network.name });
```
moshmage, line 239: you can flip this around and not need to assign stuff,

```ts
setGithub(previous => {
  ... previous,
  repositories: [
    ... await searchRepositories({networkName: network.name})
        .then(repos =>
          Promise.all(
            repos.map(repo => ({
              checked: true,
              isSaved: true,
              name: repo.githubPath.split("/")[1],
              fullName: repo.githubPath,
              hasIssues: await repositoryHasIssues(repo.githubPath)
            })
          )
        ),
    ... filtered.filter(repo =>
         !repositories.find((repoB) => repoB.fullName === repo.fullName))
  ]
});

moshmage (changes requested):

moshmage (approved):

PR #495, opened 2022-07-26
Dev 485 implement new oracles warning

MarcusviniciusLsantos:


components/card-become-council.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Button transparent className="card-close-button p-1" onClick={() => setShow(false)}>
+          <CloseIcon
+            className="text-purple"
+          />
+        </Button>
+      </div>
+      <h4 className="h4 pb-2">{t("become-council")}</h4>
+      <div className="text-gray pe-3 pb-2">
+      <span
+              dangerouslySetInnerHTML={{
```
moshmage, line 35: ?!?! _why_ ?

moshmage (changes requested):

components/card-become-council.tsx
MarcusviniciusLsantos: A solution to style a specific name within a text without having to split it in council.json

moshmage (approved):

PR #497, opened 2022-07-28
refactoring bounty list

MarcusviniciusLsantos:


components/issue-list-item.tsx
```diff
@@ -30,90 +29,119 @@ export default function IssueListItem({
...
+            </div>
+          </>
+        )}
       </div>
     );
   }
 
-  function renderReviews() {
+  function renderAvatar() {
+    if (issue?.creatorGithub)
```
moshmage, line 120: We can't have TWO profile pictures. We either show the one from github or we show the Identicon

components/bounty-status-info.tsx
```diff
@@ -8,40 +8,51 @@ interface IBountyStatusInfo {
...
   }
 
+  function renderEllipse(issueState: IssueState) {
+    if(["canceled", "closed"].includes(issueState)) return null
+    return <div className={`ellipse bg-primary me-2 ${handleColorState(true)}`} />
+  }
+
   return (
-    <span className={`status caption-small ${handleColorState()}`}>
+    <span className={`d-flex status caption-medium py-1 px-3 ${handleColorState()}`}>
```
moshmage, line 54: This can be simplified via constants (because they are constants, and we don't really need logical getters),

```tsx
const Colors = {
  open: { ellipse: 'bg-success', badge: 'bg-success-30 text-success' },
  closed: { ellipse: '', badge: 'bg-dark-gray text-white-40' }
  // ...
}
  return <Badge bg="" className="Colors[issueState].badge">
    { (Colors[issueState].badge && <div className={`ellipse bg-primary me-2 ${Colors[issueState].badge}`} /> }
    {issueState && <Translation ns="bounty" label={`status.${issueState}`} />}
  </Badge>
```

components/badge.tsx
```diff
@@ -1,22 +1,25 @@
+import { ReactElement } from "react";
 import { Badge as ReactBadge } from "react-bootstrap";
 
 interface BadgeProps {
-  label: string;
+  label?: string | ReactElement;
```
moshmage, line 5: I'm thinking you wait "either label or children" so this would be `label?: string` and not `string | ReactElement`

components/issue-list-item.tsx
```diff
@@ -30,90 +27,107 @@ export default function IssueListItem({
...
-            : renderProposals()}
-          {state?.toLowerCase() !== "draft" && (
-            <DateLabel date={issue?.createdAt} />
-          )}
-        </div>
-      );
-    } else {
-      return;
-    }
+  function handleIssueState() {
```
moshmage, line 123: repeated code -- we should be able to abstract this function into a helper and provide it with everything it needs to discern the issueType so we don't have to mantaign more than one "handleIssueState" function [which should really be called "getIssueState"]

moshmage (changes requested):

moshmage (approved):

PR #498, opened 2022-07-29
Dev 565 add a wallet address vs GitHub account validation to post put patch delete endpoints

vhcsilva:


moshmage:
@vhcsilva too many changes that aren't from this task on this PR -- You might be missing a rebase

vhcsilva:
@moshmage Rebase done

pages/api/_middleware.ts
```diff
@@ -1,23 +1,45 @@
...
 
+const UnauthorizedResponse = 
+  (reason: string) => new Response(JSON.stringify({ reason }), { status: 401, statusText: "Unauthorized" });
 
 export async function middleware(req: NextApiRequest) {
   const method = req.method;
-  const shouldCheckForToken = blackList[method]?.some(path => req.url.replace('/api/', '').search(path) > -1);
+  const endpoint = req.url.split("api/")[1];
+
+  const shouldCheckForToken = blackList[method]?.some(path => (new RegExp(`^${path}$`)).test(endpoint) );
```
moshmage, line 28: be less restrictive on the regex, and you can forgo the assignment to `endpoint = ...`, remove the `^` from this regex and we stop having to worry about the "/api/" part

pages/api/_middleware.ts
```diff
@@ -1,23 +1,45 @@
...
+type CombinedHeader = IncomingHttpHeaders & CustomHeader;
 
 const testnet = process.env.NEXT_E2E_TESTNET === "true";
 
 const blackList = {
-  GET: ['auth', 'past-events', 'seo', 'users', 'graphql'],
-  POST: ['issue']
-}
-
+  POST: ["network", "pull-request", "repos"],
```
moshmage, line 15: I'm thinking this could be reverted to how it was:

We have a list of things that _don't need_ a token and if the url is not inside then we check if a token exists. Otherwise this list will ... grow. like: gigantic. It's easier to mantaign if we just list what we allow without token

moshmage (changes requested):

contexts/authentication.tsx
```diff
@@ -157,7 +157,10 @@ export const AuthenticationProvider = ({ children }) => {
   }, [DAOService, wallet?.address]);
 
   useEffect(() => {
-    if (wallet && wallet?.address) updateWalletBalance();
+    if (!wallet?.address) return localStorage.setItem("currentWallet", "");
```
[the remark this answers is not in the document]
vhcsilva: @moshmage We can use `sessionStorage`, but we still need to set the value as "" in case of user disconnect the wallet manually using the `Disconnect Wallet` button

moshmage (approved):

PR #502, opened 2022-08-03
Dev 374 summarized bounty state page in case user is on mobile no metamask

MarcusviniciusLsantos:


components/bounty-hero.tsx
```diff
@@ -26,90 +40,95 @@ export default function BountyHero() {
...
+                  </div>
+
+                  <div>
+                    <span className="caption-small mr-1 text-white">
+                      {activeIssue?.pullRequests?.length || 0}
+                    </span>
+                    <span className="caption-small text-white-40 text-uppercase">
+                      {activeIssue?.pullRequests?.length === 1
+                        ? t("info.pull-requests_one")
+                        : t("info.pull-requests_other")}
```
vhcsilva, line 100: No need to do that, i18next can handle plurals.

https://www.i18next.com/translation-function/plurals

components/bounty-hero.tsx
```diff
@@ -26,90 +40,95 @@ export default function BountyHero() {
...
+                  </div>
+
+                  <div>
+                    <span className="caption-small mr-1 text-white">
+                      {activeIssue?.mergeProposals?.length || 0}
+                    </span>
+                    <span className="caption-small text-white-40 text-uppercase">
+                      {activeIssue?.mergeProposals?.length === 1
+                        ? t("info.proposals_one")
+                        : t("info.proposals_other")}
```
vhcsilva, line 111: Use i18next too to show plurals

contexts/issue.tsx
```diff
@@ -114,9 +114,9 @@ export const IssueProvider: React.FC = function ({ children }) {
     [activeNetwork, query?.repoId, query?.id, user?.accessToken]);
 
   const getNetworkIssue = useCallback(async () => {
-    if (!activeIssue?.contractId || !DAOService?.network?.contractAddress)
+    if ((!activeIssue?.contractId || !DAOService?.network?.contractAddress) || !wallet?.address)
```
vhcsilva, line 117: An user can access web-network from PC and have metamask installed, but don't connected a wallet yet, we still need to load the blockchain data in that case.

I think that for the mobile issue the condition `!DAOService?.network?.contractAddress` is enough, because `contractAddress` on `DAOService.network` will only exists after the contract loads, and this will only happen if metamask is installed.

vhcsilva (changes requested):

vhcsilva (approved):

PR #503, opened 2022-08-03
DEV-656-hide-become-a-council-card-if-user-already-is-a-council-member

MarcusviniciusLsantos:


vhcsilva (approved):

components/oracle.tsx
```diff
@@ -105,7 +107,7 @@ export default function Oracle({ children }) {
       <div className="container p-footer">
         <div className="row justify-content-center">
           <div className="col-md-10 mt-2">
-            <CardBecomeCouncil />
+            {!wallet?.isCouncil && <CardBecomeCouncil />}
```
moshmage, line 110: Won't this also show if you don't connect a wallet ?

MarcusviniciusLsantos: not. It displays even without linked wallet

PR #505, opened 2022-08-03
Dev 654 show address truncated when a bounty was created without a GitHub account

MarcusviniciusLsantos:


components/not-mergeable-modal.tsx
```diff
@@ -43,7 +43,7 @@ export default function NotMergeableModal({
   const { activeIssue, networkIssue } = useIssue();
 
 
-  const isIssueOwner = activeIssue?.creatorGithub === user?.login;
+  const isIssueOwner = activeIssue?.creatorAddress.toLowerCase() === wallet?.address.toLowerCase();
```
vhcsilva, line 46:
```suggestion
  const isIssueOwner = !!activeIssue && activeIssue?.creatorAddress?.toLowerCase() === wallet?.address?.toLowerCase();
```
I think without optional chaining it can throw "reading of undefined"  and we need to add a check for `activeIssue` because it can be undefined until the context loads.

vhcsilva (changes requested):

helpers/formatAddress.ts
```diff
@@ -0,0 +1 @@
+export const formatAddress = (address:  string, joiner = "...") => [address.slice(0, 4), address.slice(-4)].join(joiner)
```
vhcsilva, line 1: We can delete this file, as it is not used anymore

vhcsilva (changes requested):

moshmage (approved):

vhcsilva (approved):

PR #506, opened 2022-08-03
Dev 601 separate profile balances page

vhcsilva:


moshmage (approved):
`components/oracle-actions` and `...-delegate` could be refactored to not re-call useAuthentication and instead use the values available from `oracles.tsx` but other than that, lgtm

PR #510, opened 2022-08-04
Dev 674 the distribution modal for the funding bounty type is broken

vhcsilva:


moshmage (approved):
I keep forgetting that function exists! Nicely done :)

PR #517, opened 2022-08-09
Dev 676 add date range filter to the payments page

MarcusviniciusLsantos:


components/profile/payments-list.tsx
```diff
@@ -1,16 +1,74 @@
...
   payments: Payment[];
 }
 
 // TODO: Add InfiniteScroll and pagination
 export default function PaymentsList({ payments } : PaymentsListProps) {
+  const { push } = useRouter();
+  const { t } = useTranslation(["common", "profile", "bounty"]);
+  const { getURLWithNetwork } = useNetworkTheme();
+
+  function PaymentItem({ ammount, issue, transactionHash, id} : Payment) {
```
moshmage, line 24: filename is called "payments list", it should only contain ... payments-list. there was a file named "payment-item" because we could want to use that same component elsewhere :)

moshmage (changes requested):

components/profile/payments-list.tsx
MarcusviniciusLsantos: i had a problem to use the translations in the payment-item component this was the solution that solved this problem. But I can investigate further and see other solutions like passing the translation

components/profile/payment-item.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { push } = useRouter();
+
+  const { getURLWithNetwork } = useNetworkTheme();
+
+  const CONTAINER_CLASSES = [
+    "justify-content-between align-items-center bg-transparent",
+    "border border-dark-gray border-radius-8 mb-2 py-3 px-4"
+  ];
+
+  function handleClick() {
```
moshmage, line 22: handle-click should be handled by the parent component, this should be a component argument

moshmage (changes requested):

moshmage (approved):

PR #519, opened 2022-08-10
Dev 681 review create bounty does not show select token

MarcusviniciusLsantos:


components/create-bounty-modal.tsx
```diff
@@ -306,26 +288,26 @@ export default function CreateBountyModal() {
...
-                <GithubInfo
-                  parent="list"
-                  variant="repository"
-                  label={repository?.path}
-                  simpleDisabled={true}
-                />
+              <GithubInfo
+                parent="list"
+                variant="repository"
+                label={repository?.path.length <= 20 ? repository?.path : repository?.path.slice(0,20)+"..."}
```
moshmage, line 298: this should be done via css

moshmage (changes requested):

components/create-bounty-modal.tsx
MarcusviniciusLsantos: in this situation the component cannot have the dimension to perform the truncate. I put the truncate in the div above, but it didn't have such a nice visual effect. So I put this slice to improve the viewing experience

moshmage (approved):

PR #526, opened 2022-08-12
Dev 683 backend does not allow two networks to the same GitHub user but this should be allowed

vhcsilva:


pages/api/network/index.ts
```diff
@@ -47,14 +47,14 @@ async function post(req: NextApiRequest, res: NextApiResponse) {
...
-
-    if (!user) return res.status(403).json("Invalid user provided");
+    // const user = await Database.user.findOne({
+    //   where: {
+    //     githubLogin,
+    //     address: creator.toLowerCase()
+    //   }
+    // });
+
+    // if (!user) return res.status(403).json("Invalid user provided");
```
clarkjoao, line 57: [sugestion] Wouldn't it be better to delete the code?

pages/api/network/index.ts
```diff
@@ -168,14 +168,14 @@ async function put(req: NextApiRequest, res: NextApiResponse) {
...
     const isAdminOverriding = !!override;
 
-    const user = await Database.user.findOne({
-      where: {
-        githubLogin,
-        address: creator.toLowerCase()
-      }
-    });
+    // const user = await Database.user.findOne({
+    //   where: {
```
clarkjoao, line 172: same to here

clarkjoao (approved):

MarcusviniciusLsantos (approved):

PR #532, opened 2022-08-16
Dev 690 add funded amount column to issue table and update it using bounty funded event

MarcusviniciusLsantos:


components/bounty/funding-section/fund-modal.tsx
```diff
@@ -11,18 +11,22 @@ import Modal from "components/modal";
...
 import useERC20 from "x-hooks/use-erc20";
 
 import { Amount, CaptionMedium, RowWithTwoColumns } from "./minimals";
 
 export default function FundModal({
   show = false,
-  onCloseClick
+  onCloseClick,
+  repoId,
+  ghId
```
vhcsilva, line 29: We are using the `issue` context in this component, so we don't need to pass theses ids by props

vhcsilva (changes requested):

clarkjoao (approved):

vhcsilva (approved):

PR #536, opened 2022-08-18
Dev 695 webapp must request past events to microservice

clarkjoao:


vhcsilva (approved):

MarcusviniciusLsantos (approved):

scripts/deploy-v2.js
```diff
@@ -92,7 +92,7 @@ const networks = {
 
 
 async function main() {
-
+  console.log({ host: process.env.NEXT_DB_DATABASE, ...process.env })
```
moshmage, line 95: use `console.table`

clarkjoao: I will remove, it was only to try see the envs in Jenkins

components/price-conversor-modal.tsx
```diff
@@ -53,9 +53,13 @@ export default function PriceConversorModal({
 
   useEffect(()=>{
     const currencyList = settings?.currency?.conversionList || defaultValue;
-    const opt = currencyList.map(currency=>({value: currency?.value, label: currency?.label}))
-    setOptions(opt)
-    handlerChange(opt[0])
+    console.log(currencyList)
```
moshmage, line 56: lingering console-log

moshmage (approved):
lgtm

moshmage:
@clarkjoao I just noticed that since most of the events will run at the same time, we might get conflicting states (pex, creating and immedeatly canceling a bounty would trigger the action to go hay wire) -- we need to offset the timers by 1 at least

(every 30mins, 31mins, 32mins, etc)

PR #546, opened 2022-08-23
Dev 547 use a database for dapp configurations instead of relying on env files

vhcsilva:


pages/[network]/proposal.tsx
```diff
@@ -63,6 +63,8 @@ export default function PageProposal() {
         }));
       })
       .catch(error => {
+        if (error?.code === 4001) return;
```
Comment (moshmage) [inline pages/[network]/proposal.tsx:66]: what error is this? :o

Comment (vhcsilva) [reply pages/[network]/proposal.tsx:66]: @moshmage is "rejected transaction", we don't need to show a toast if user rejects the transaction

components/transaction-modal.tsx
```diff
@@ -99,7 +101,7 @@ export default function TransactionModal({
   }
 
   function getEtherScanHref(tx: string) {
-    return `//${publicRuntimeConfig?.metaMask?.blockScanUrl}/tx/${tx}`;
+    return `//${settings?.urls?.blockScan}/tx/${tx}`;
```
moshmage, line 104: we found out that this hardcoded `tx` might screw _other_ blockscanners -- best to _not_ include that hardcoded string and just have the value on the database have it

contexts/index.tsx
```diff
@@ -9,26 +9,28 @@ import { IssueProvider } from "contexts/issue";
...
-                  <CreateBountyModal />
-                  {children}
-                </IssueProvider>
-              </ReposProvider>
-            </ApplicationContextProvider>
-          </NetworkSettingsProvider>
-        </AuthenticationProvider>
-      </NetworkProvider>
-    </DAOContextProvider>
+    <SettingsProvider>
```
moshmage, line 16: note: We will have to start thinking on having a `MetaProvider` component that includes all the context-providers because this index.tsx is getting ridiculous :P

vhcsilva: @moshmage this file exports the `RootProviders` component which is used in `_app.tsx`

helpers/api/proposal/read-refused.ts
```diff
@@ -51,16 +47,7 @@ export default async function readProposalRefused(events, network: Network_v2, c
           });
 
           if (proposal) {
-            if (network.contractAddress === publicRuntimeConfig?.contract?.address) {
```
moshmage, line 54: ?

moshmage (approved):
lgtm, take care with the deletion of the twitter thingies

helpers/api/proposal/read-refused.ts
vhcsilva: @moshmage twitter integration will be made by webnetwork-events project

db/migrations/20220816114655-create-settings-table.js
```diff
@@ -0,0 +1,63 @@
...
+      key: {
+        type: Sequelize.STRING,
+        allowNull: false
+      },
+      value: {
+        type: Sequelize.STRING,
+        allowNull: false
+      },
+      type: {
+        type: Sequelize.ENUM("string", "boolean", "number", "json"),
```
moshmage, line 27: ??! _why_? xD

moshmage (commented):
still lgtm

PR #551, opened 2022-08-25
feat: reset gh account

clarkjoao:


moshmage:
@clarkjoao is this deployed into any env ?

vhcsilva (approved):

PR #555, opened 2022-09-01
Dev 705 match web network to dappkit 2 0 9

MarcusviniciusLsantos:


components/change-token-modal.tsx
```diff
@@ -45,12 +45,11 @@ export default function ChangeTokenModal({
...
-      }
+      } 
 
-      const erc20 = await DAOService.loadERC20(address);
-
-      setName(await erc20.name());
-      setSymbol(await erc20.symbol());
+      const token = await DAOService.loadERC20(address)
+      setName(await token.name());
+      setSymbol(await token.symbol());
```
vhcsilva, line 52: `DAOService` has the `getERC20TokenData` method that can resume these lines

components/custom-network/token-configuration.tsx
```diff
@@ -198,38 +183,55 @@ export default function TokenConfiguration({
...
 
-      <DeployNFTModal 
+      <div className="row">
+          <>
+            <MultipleTokensDropdown
+              label={t("select-multiple.allowed-transactional-tokens")}
+              description={t("select-multiple.add-transactional-tokens")}
+              addToken={addTransactionalToken}
+              tokens={handleEmptyTokens(allowedTransactionalTokens)}
+              canAddToken={false}
```
moshmage, line 214: shouldn't this depend on the user wallet instead of hardcoded to false ?

components/profile/my-network-settings.tsx
```diff
@@ -425,10 +431,27 @@ export default function MyNetworkSettings({ network, updateEditingNetwork } : My
             value={settings?.treasury?.closeFee?.value}
             error={settings?.treasury?.closeFee?.validated === false}
             onChange={fields.closeFee.setter}
+            disabled={true}
```
moshmage, line 434: shouldn't this depend on the users wallet and if it is a network governor?

components/custom-network/token-configuration.tsx
MarcusviniciusLsantos: not in this situation because it is creating the network and can only use allowed tokens

components/profile/my-network-settings.tsx
MarcusviniciusLsantos: I created a section to control this situation and left this input as default disabled for the user to see that it exists but he has no control

components/tokens-settings.tsx
```diff
@@ -43,34 +43,40 @@ export default function TokensSettings({
...
+    }
 
     if (!isGovernorRegistry) {
-      getTokens()
-        .then((tokens) => {
-          setAllowedTransactionalTokens(tokens.filter((token) => token.isTransactional === true));
-          setAllowedRewardTokens(tokens.filter((token) => token.isTransactional === false));
-        })
-        .catch((err) => console.log("error to get tokens:", err));
+      DAOService.getAllowedTokens().then((allowedTokens) => {
```
moshmage, line 57: this can be made into a function and reused. no point in having the exact same logic twice

services/dao-service.ts
```diff
@@ -96,6 +99,48 @@ export default class DAO {
     return token;
   }
 
+
+  transactionHandler(event: PromiEvent<TransactionReceipt | Contract>,
+    resolve: ResolveReject,
```
moshmage, line 104: these arguments are not respecting code-style-spacing , they should be aligned with "event"

services/dao-service.ts
```diff
@@ -96,6 +99,48 @@ export default class DAO {
...
+    reject: ResolveReject,
+    debug = false) {
+
+    let _iban: string;let tries = 1;
```
moshmage, line 108: assign variables in their own line, please

services/dao-service.ts
```diff
@@ -257,13 +302,19 @@ export default class DAO {
   async addAllowedTokens(addresses: string[], isTransactional: boolean) {
     if (!this.registry) await this.loadRegistry();
 
-    return this.registry.addAllowedTokens(addresses, isTransactional);
+    this.web3Connection.options.customTransactionHandler = this.transactionHandler.bind(this) 
+
+    return this.registry.addAllowedTokens(addresses, isTransactional)
+    .finally(() => this.web3Connection.options.customTransactionHandler = null)
```
moshmage, line 308: not matching code-style-spacing, this should be alligned with either "registry" or "addAllowedTokens"

moshmage (changes requested):

db/migrations/2022083100590000-remove-unique-addres-to-tokens.js
```diff
@@ -0,0 +1,20 @@
...
+  async up (queryInterface, Sequelize) {
+    await queryInterface.changeColumn('tokens', 'address', {
+        type: Sequelize.STRING,
+        allowNull: false,
+        unique: false
+      });
+    await queryInterface.removeConstraint('tokens', 'tokens_address_key')
+  },
+
+  async down (queryInterface, Sequelize) {
```
moshmage, line 13: aren't we missing "addConstraints" ?

pages/api/registry/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+
+  return res.status(200).json("Update allowed tokens in database");
+}
+
+async function UpdateAllowedTokens(req: NextApiRequest,
+                                   res: NextApiResponse) {
+  switch (req.method.toLowerCase()) {
```
moshmage, line 97: You're missing a POST method to _change_ the database. GET should only return, POST can change.

You also need to make sure that the person who's making this change to the database actually _can_ make this change - otherwise _anyone_ would be able to make this change.

db/migrations/2022083100590000-remove-unique-addres-to-tokens.js
MarcusviniciusLsantos: this table was created as address unique, I had to remove it to use isTransactional logic

pages/api/registry/index.ts
moshmage: @MarcusviniciusLsantos like we spoke, this script belongs on the webnetwork-events

x-hooks/use-api.tsx
```diff
@@ -441,7 +441,7 @@ export default function useApi() {
 
   async function updateAllowedTokens() {
     return client
-      .get(`/registry`)
+      .post(`/registry`)
```
moshmage, line 444: like we spoke, we should query the webnetwork-events instead

moshmage (changes requested):

db/migrations/2022083100430000-add-isTransactional-to-tokens.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface
+    .addColumn("tokens", "isTransactional", {
```
vhcsilva, line 6: I think that would be better to have `isTransactional` column in the `network_tokens` table, because a token can be used as transactional in one network and as reward in other. By moving the column we don't need to duplicate token information in the database.

Do you agree @moshmage ?

components/profile/registry-governor-settings.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export default function RegistryGovernorSettings() {
+  const { t } = useTranslation(["common", "custom-network"]);
+  const { dispatch } = useContext(ApplicationContext);
+  const { fields, settings } = useNetworkSettings();
+  const {  activeNetwork, updateActiveNetwork } = useNetwork();
+  const { service: DAOService } = useDAO();
+  const txWindow = useTransactions();
+
+  async function handleFeeSettings(): Promise<TransactionReceipt | Error> {
```
vhcsilva, line 34: This function can be moved to the `useBepro` hook

components/custom-network/network-parameter-input.tsx
```diff
@@ -32,7 +32,6 @@ export default function NetworkParameterInput({
         placeholder={"0"}
         thousandSeparator
         onValueChange={handleChange}
-        decimalScale={decimals}
```
moshmage, line 31: ?

deploy-v2.js
```diff
@@ -0,0 +1,162 @@
+const {
+  Web3Connection,
+  ERC20,
+  BountyToken,
+  Network_v2,
+  NetworkFactoryV2,
+  toSmartContractDecimals,
+  Network_Registry
+} = require("@taikai/dappkit");
+const { nativeZeroAddress } = require("@taikai/dappkit/dist/src/utils/constants");
```
moshmage, line 10: You can import this from `@taikai/dappkit` by importing `import {Defaults} from '@taikai/dappkit'` and then use it as `Defaults.nativeZeroAddress`

deploy-v2.js
```diff
@@ -0,0 +1,162 @@
...
+const fs = require("fs");
+
+const cap = toSmartContractDecimals("50000000", 18);
+
+const connection = new Web3Connection({
+  web3Host: "http://127.0.0.1:8545",
+  privateKey:
+    "0x40a674d5327e433ccb291408db12e890a806bfa49916e2fd948770b246d6a871",
```
moshmage, line 18: hardcoded pvtkeys can't be a thing -- lets all hope this is a demo private key and _not your own_

deploy-v2.js
```diff
@@ -0,0 +1,162 @@
...
+  toSmartContractDecimals,
+  Network_Registry
+} = require("@taikai/dappkit");
+const { nativeZeroAddress } = require("@taikai/dappkit/dist/src/utils/constants");
+const fs = require("fs");
+
+const cap = toSmartContractDecimals("50000000", 18);
+
+const connection = new Web3Connection({
+  web3Host: "http://127.0.0.1:8545",
```
moshmage, line 16: same for web3host, these need to be params / env vars

deploy-v2.js
```diff
@@ -0,0 +1,162 @@
...
+  console.log("Done.");
+
+  console.log("Changing parameters...");
+  await networkObj.changeDraftTime(61);
+  await networkObj.changeDisputableTime(120);
+  await networkObj.changeCouncilAmount(101000);
+  console.log("Done.");
+
+  const address1 = "0xf3ad222f9Aa841C240022b73cA228A71a92fF82D";
+  const address2 = "0xC3869b5a2D0EA422a007a2503C0d616C6a8d6B63";
```
moshmage, line 129: make this addresses and array, and then iterate over them,
```ts
await Promise.all(
  ['address', 'address']
    .map(address => {
      await settler.transferTokenAmount(address, 10000000)
      await transactional.transferTokenAmount(address, 10000000);
      await reward.transferTokenAmount(address, 10000000);
    }):
```

PS: Make sure to only do this if we haven't provided these values, because if we have provided these contract addresses (reward, transactional, settler) then these are already existing coins and most likely _the person running the script_ won't want to transfer more things.

moshmage (changes requested):
Take care with 
- repeated .env files along 
- repeated deploy-scripts _outside_ the scripts folder. 

I'm sure these are testing scripts, not sure they should be committed.

LOOKOUT with plain-text private keys on the repo. make sure
1) you don't commit those
2) that if you do, it's a test one

moshmage (approved):

PR #563, opened 2022-09-07
Dev 808 the amount field has no limit on decimal places

vhcsilva:


scripts/settings/generate-types.js
```diff
@@ -0,0 +1,49 @@
+require("dotenv").config();
```
moshmage, line 1: hmm.. I see this as a tool that you have _for yourself_ so as to not write these types by hand -- but these types shouldn't be automated, mainly because the database _cannot_ control our source code - it should be the other way around

moshmage: _ ... I didn't notice this was in draft -- Ignore me_

moshmage (approved):

PR #574, opened 2022-09-15
Fix dev 712 change the network creation flow

vhcsilva:


helpers/api/registry/read-registered.ts
```diff
@@ -0,0 +1,59 @@
+import * as Logging from "@scripts/logging"
```
moshmage, line 1: ? I think this file belongs on the webevents-api, webnetworks no longer uses event reading -- we do that on the webnetwork-events 

If you need to also read that event via "?fromBlock=" to parse a user event, simply add the route for it.

moshmage (changes requested):

moshmage (approved):

PR #582, opened 2022-09-24
Dev 850 remove treasury info from network creation page

clarkjoao:


pages/api/search/issues/index.ts
```diff
@@ -97,13 +98,21 @@ async function get(req: NextApiRequest, res: NextApiResponse) {
     { association: "repository" },
     { association: "token" }
   ];
-
+  const sortBy = String(req?.query?.sortBy).split(',');
```
moshmage, line 101: Coulnd't you have replaced `,` for `+` and be done with it?

```
const sortBy = String(req?.query?.sortyBy).replace(',','+'); // word+word+word
```

pages/api/search/issues/index.ts
```diff
@@ -121,10 +130,17 @@ async function get(req: NextApiRequest, res: NextApiResponse) {
       currentPage: +paginatedData.page
     });
   } else {
+
+    const fields = sortBy?.reduce((prev, current, index) => {
```
moshmage, line 134: in either way, you can pull the logic up on the function and not have to repeat it

moshmage (changes requested):

moshmage (approved):

PR #583, opened 2022-09-26
Dev 855 change navbar to remove bv token

MarcusviniciusLsantos:


moshmage (approved):

moshmage:
@MarcusviniciusLsantos only thing missing is change the `/profile/oracles/` to `/profile/bepro-votes`

PR #586, opened 2022-09-27
Dev 802 network create fields arent saveds

clarkjoao:


contexts/network-settings.tsx
```diff
@@ -64,120 +62,199 @@ export const NetworkSettingsProvider = ({ children }) => {
...
+    newState.tokens.validated = tokensValidated;
+    newState.isSettingsValidated = [tokensLockedValidate,
+                                    detailsValidate,
+                                    githubValidate,
+                                    settingsValidated,
+                                    tokensValidated,
+    ].every(condtion=> condtion)
+
+    if(detailsValidate)
+      localStorage.setItem(`${localStorageKey}:${wallet.address}`,
```
vhcsilva, line 170: Are we clearing this information when the user disconnect ? @clarkjoao

clarkjoao: No, only past 2 days or after create network.

vhcsilva: As we are clearing other data from localStorage on disconnect, i think we can do it for network settings data too

clarkjoao: I did a double check, and when we disconnect we are cleaning every items in localstorage. 🙂

Screenshot 2022-09-28 at 21 01 28

moshmage (changes requested):
decrease the debounce time from 1000 (1sec) to 500ms (half a second)

moshmage:
@clarkjoao check the conflicts

PR #587, opened 2022-09-27
#864 Update Deploy v2 Script - Production Flag

hvasconcelos:


scripts/deploy-v2.js
```diff
@@ -210,7 +221,8 @@ async function main() {
 
 
     console.log(`Deploying Network_V2 With Registry...`);
-    const registryReceipt = await Deployer(NetworkRegistry, [networkToken.contractAddress, 1000, ownerAddress, 10000]);
+    const treasuryAddress = argv.treasuryAddress ? argv.treasuryAddress: ownerAddress;
+    const registryReceipt = await Deployer(NetworkRegistry, [networkToken.contractAddress, 1000, treasuryAddress, 10000]);
```
moshmage, line 225: missing arguments for "network creation value", because I'm thinking we dont want that to be `1000`

scripts/deploy-v2.js
```diff
@@ -242,36 +262,41 @@ async function main() {
...
+        }),
+        updateTokens({
+          name: await bountyTransactional.name(),
+          symbol: await bountyTransactional.symbol(),
+          isTransactional: true,
+          address: bountyTransactional.contractAddress
+        })
+      ]);
+    } else {
+      console.log('Skiping Database Save');
```
moshmage, line 297: If we're not saving this information to the database, _how_ are we to know these values when the application queries for them ?

moshmage (changes requested):

scripts/deploy-v2.js
hvasconcelos: On production, we can change adjust that later on the Network registry.

scripts/deploy-v2.js
hvasconcelos: They are coming from environment variables right ?! At least was what you said.

moshmage (approved):

PR #588, opened 2022-09-27
DEV-853-fix-the-flow-of-visualization-of-the-values-received-by-the-delegate

MarcusviniciusLsantos:


moshmage (approved):
I'm not sure about the `Number()` use here as it _might_ round value when it goes from string to number -- but I'm willing to live wit this since @vhcsilva story should also take care of this

PR #590, opened 2022-09-28
settings should account for process.env because _reasons_

moshmage:
The publicRuntimeConfig should
1. match the new settings
2. be intertwined with the settings
3. but overwritten if database settings exist

---

To achieve this, these changes were made

- add publicRuntimeConfig to settings
- add uri to closeBounty
- recreate public runtime config

effectively, if we save our public settings to the database nothing alters, otherwise it will use the (public) values we have in our `.env`

---

This branch is a in-between and should be only merged on 2.1 when we remove the settings database from existance. This was a messed up idea, and we have to revert it. my bad.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -9,6 +9,68 @@ const publicRuntimeConfig = {
     api: process.env.NEXT_PUBLIC_API_HOST || "http://localhost:3000",
     home: process.env.NEXT_PUBLIC_HOME_URL || "http://localhost:3000",
     events: process.env.NEXT_PUBLIC_EVENTS_API || "http://localhost:3334",
+    ipfs: process.env.NEXT_PUBLIC_IPFS_BASE,
+    nft: process.env.NEXT_PUBLIC_NFT_URI,
+    web3Provider: process.env.NEXT_PUBLIC_WEB3_CONNECTION,
+    blockScan: process.env.NEXT_PUBLIC_BLOCKSCAN_LINK
+  },
+  contracts:{
+    settlerToken: process.env.NEXT_PUBLIC_SETTLER_ADDRESS,
+    network: process.env.NEXT_PUBLIC_CONTRACT_ADDRESS,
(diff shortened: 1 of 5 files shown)


next.config.js
```diff
@@ -9,6 +9,68 @@ const publicRuntimeConfig = {
...
+    42: 'kovan',
+    3: 'ropsten',
+    4: 'rinkeby',
+    5: 'goerli',
+    1285: 'moonriver',
+    1337: 'localhost',
+    1500: 'seneca',
+    1501: 'afrodite',
+    1502: 'irene',
+    1503: 'iris'
```
vhcsilva, line 73:
```suggestion
    1503: 'iris',
    1504: 'diogenes',
    1505: 'aurelius'
```

next.config.js
```diff
@@ -9,6 +9,68 @@ const publicRuntimeConfig = {
...
+    defaultToken: process.env.NEXT_PUBLIC_CURRENCY_ID,
+    conversionList: process.env.NEXT_PUBLIC_CURRENCY_VSLIST,
+  },
+  defaultNetworkConfig: {
+    networkName: process.env.NEXT_PUBLIC_DEFAULT_NETWORK_NAME,
+    allowCustomTokens: process.env.NEXT_PUBLIC_ALLOW_CUSTOM_TOKENS,
+    adminWalletAddress: process.env.NEXT_PUBLIC_ADMIN_WALLET_ADDRESS,
+  },
+  nftUri: process.env.NEXT_PUBLIC_NFT_URI,
+  web3ProviderConnection: process.env.NEXT_PUBLIC_WEB3_CONNECTION,
```
vhcsilva, line 36: We already have these envs in the urls property

vhcsilva (changes requested):

next.config.js
moshmage: forgot to remove them

vhcsilva (approved):

clarkjoao (approved):

moshmage:
closing because outdated

PR #592, opened 2022-09-29
Dev 807 front end is rounding values when it has many decimal places

vhcsilva:


components/council-layout.tsx
```diff
@@ -67,7 +67,7 @@ export default function CouncilLayout({ children }) {
         currency: "BEPRO",
       },
       {
-        value: onNetwork,
+        value: onNetwork.toNumber(),
```
moshmage, line 70: shouldn't this be to fixed?

services/dao-service.ts
```diff
@@ -184,29 +187,30 @@ export default class DAO {
...
       case 'staked':
-        n = await this.network.totalNetworkToken();
+        n = new BigNumber(await this.network.totalNetworkToken()).decimalPlaces(decimals);
         break;
       }
-      
-      return n;
     } catch (error) {
-      return 0;
+      console.debug("Failed to getBalance", error);
```
moshmage, line 207: add the type of balance we failed to get so we can debug it properly

moshmage (changes requested):

moshmage (approved):

PR #613, opened 2022-10-12
DEV 838 When a huge number is typed into Lock's, Approve and Get buttons are broken

MarcusviniciusLsantos:


components/oracles-actions.tsx
```diff
@@ -78,7 +78,7 @@ function OraclesActions({
                token: activeNetwork?.networkToken?.symbol 
              }),
       label: t("my-oracles:actions.lock.get-amount-oracles", {
-        amount: formatNumberToNScale(tokenAmount),
+        amount: new BigNumber(tokenAmount).gte(1e18) ? null : formatNumberToNScale(tokenAmount),
```
moshmage, line 81: ?!?!?

PR #621, opened 2022-10-14
chore: fix states

moshmage:
An overly complicated logic was used when reducers were available; as such, we need to fix this before 2.2 comes around. 

The main Idea is to use `SimpleAction` class and extend it to the needs. This `SimpleAction` extension is then loaded in by adding it to the `loadApplicationStateReducers()` list via `addReducer()`

All of this makes use of reacts `useReducer` and `useContext`;

Reading state
```ts
const {state} = useContext(AppStateContext)

console.log(state.prop);
```

Changing state
```ts
// import SimpleAction
const {dispatch} = useContext(AppStateContext);

dispatch(new SimpleAction(0, 'prop').update('some-state'));
```

Base logic

```ts
export abstract class Actor<T = any> {
  readonly id: AppStateReduceId;
  abstract update(payload: T): {id: AppStateReduceId, payload: T};
}

export abstract class Action<T = any> {
  readonly id: AppStateReduceId;
  readonly stateKey?: string;

  abstract reducer(state: State, payload: T): State;
}

export class SimpleActor<T = any> implements Actor<T> {
  constructor(readonly id) {}

  update(payload: T): { id: AppStateReduceId; payload: T } {
    return {id: this.id, payload};
  }
}

export class SimpleAction<T = any> extends SimpleActor<T> implements Action<T> {
  constructor(readonly id: AppStateReduceId, readonly stateKey: keyof State) {
    super(id)
  }

  reducer(state: State, payload: T): State {
    console.debug(`updating`, this.stateKey, payload);
    return {...state, [this.stateKey]: payload};
  }
}

export type XReducerAction<T> = { id: AppStateReduceId, payload: T };
export type XReducer<T> = Actor<T> & Action<T>;
```

obviously, this can be turned into awesomeness. You can check `change-transaction` or `change-toaster` for examples with some liberties.


contexts/app-state.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {toastError} from "./reducers/add-toast";
+import {changeChain} from "./reducers/change-chain";
+
+const appState: AppState = {
+  state: {
+    Settings: null,
+    Service: null,
+    loading: null,
+    currentUser: null,
+    connectedChain: null,
```
moshmage, line 16: rename to "activeChain"

components/transactions-list.tsx
```diff
@@ -102,7 +102,7 @@ export default function TransactionsList({
   function clearTransactionsList() {
     localStorage.setItem(`bepro.transactions:${wallet?.address?.toLowerCase()}`, "[]");
 
-    dispatch(clearTransactions());
+    dispatch(updateTx.update([]));
```
moshmage, line 105: call `changeTx.update([])` instead

components/transactions-state-indicator.tsx
```diff
@@ -6,38 +6,66 @@ import TransactionIcon from "assets/icons/transaction";
...
 import { Transaction } from "interfaces/transaction";
+import {AppStateContext} from "../contexts/app-state";
+import {WinStorage} from "../services/win-storage";
+import {changeTxList} from "../contexts/reducers/change-tx-list";
 
 export default function TransactionsStateIndicator() {
-  const {
-    state: { myTransactions }
-  } = useContext(ApplicationContext);
+  const {state: { transactions, currentUser: { walletAddress } }, dispatch} = useContext(AppStateContext);
```
moshmage, line 16: `currentUser` is null on state start, can't be destroyed

contexts/application.tsx
```diff
@@ -73,123 +67,16 @@ const defaultState: GlobalState = {
 
 export const ApplicationContext = createContext<GlobalState>(defaultState);
 
-const cheatAddress = "";
-let waitingForTx = null;
-
 export default function ApplicationContextProvider({ children }) {
-  const [state, dispatch] = useReducer(mainReducer, defaultState.state);
-  const [txListener, setTxListener] = useState<NodeJS.Timeout | undefined>();
+  const [, dispatch] = useReducer(() => {}, null);
```
moshmage, line 71: todo: delete this file once `ApplicationContext` is no longer used

contexts/reducers/change-service-starting.ts
```diff
@@ -0,0 +1,19 @@
+import {ServiceState, State} from "../../interfaces/application-state";
```
moshmage, line 1: superfulous

contexts/reducers/change-service-starting.ts
```diff
@@ -0,0 +1,19 @@
...
+import {Action, SimpleActor} from "./reducer";
+
+export class ChangeServiceProp extends SimpleActor<boolean> implements Action<boolean> {
+
+  constructor(readonly id,
+              readonly prop: keyof ServiceState) {
+    super(id);
```
moshmage, line 8: if we add `AppStateIdThing.Service` to this we can then refer to `super.reduce()` and pass on `transformed` enabling us to use the base implementation with the debugging abilities

components/status-bar.tsx
```diff
@@ -5,17 +5,14 @@ import { useTranslation } from "next-i18next";
...
 import useApi from "x-hooks/use-api";
+import {AppStateContext} from "../contexts/app-state";
 
 
 export default function StatusBar() {
-  const {
-    dispatch,
-    state: { microServiceReady }
-  } = useContext(ApplicationContext);
+  const {dispatch, state: { Service }} = useContext(AppStateContext);
```
clarkjoao, line 15: Would be better exporting the AppContext instanced? like the others contexts?

x-hooks/use-bounty.tsx
```diff
@@ -0,0 +1,203 @@
...
+      })
+      .then(comments => {
+        dispatch(changeCurrentBountyComments(comments));
+        dispatch(changeSpinners.update({bountyDatabase: false}))
+      });
+
+  }
+
+  function getChainBounty(force = false) {
+    if (!state.Service?.active || !state.Service?.network || !state.currentBounty?.data?.contractId || state.spinners?.bountyChain)
```
vhcsilva, line 89: `!state.currentBounty?.data?.contractId` can't be used because `contractId` of a bounty can be `0`

vhcsilva, line 89: `!state.currentBounty?.data?.contractId` can't be used because `contractId` of a bounty can be `0`

```suggestion
    if (!state.Service?.active || !state.Service?.network || state.currentBounty?.data?.contractId === null || state.currentBounty?.data?.contractId === undefined || state.spinners?.bountyChain)
```

x-hooks/use-bounty.tsx
```diff
@@ -0,0 +1,203 @@
...
+        // dispatch(changeCurrentBountyDataProposals(proposals));
+        return proposals;
+      })
+      .catch(e => {
+        console.error(`Failed to get extended proposals`, e);
+        return;
+      });
+  }
+
+  function getExtendedPullRequestsForCurrentBounty() {
```
vhcsilva, line 163: Could `getExtendedProposalsForCurrentBounty` and `getExtendedPullRequestsForCurrentBounty`  be used in `getChainBounty` and `getDatabaseBounty` ?

x-hooks/use-authentication.tsx
```diff
@@ -0,0 +1,209 @@
...
+          return;
+
+        lastUrl.value = asPath;
+
+        return signedIn ? signIn('github', {callbackUrl: `${URL_BASE}${asPath}`}) : null;
+      })
+  }
+
+  function validateGhAndWallet() {
+    if (!state.currentUser?.walletAddress || !(session?.data?.user as any)?.login || state.spinners?.matching)
```
vhcsilva, line 120:
```suggestion
// Use CustomSession interface for session data
    if (!state.currentUser?.walletAddress || !(session?.data?.user as any)?.login || state.spinners?.matching)
```

x-hooks/use-authentication.tsx
```diff
@@ -0,0 +1,209 @@
...
+        else if (user.githubLogin && userLogin)
+          dispatch(changeCurrentUserMatch(userLogin === user.githubLogin &&
+            (walletAddress ? walletAddress === user.address : true)));
+      })
+      .finally(() => {
+        dispatch(changeSpinners.update({matching: false}));
+      })
+  }
+
+  function listenToAccountsChanged() {
```
vhcsilva, line 141: The way this functions is being used creates the following side effect: if the user didn't clicked on the "Connect Wallet" button yet and change the wallet on metamask it will automatically connect to the app. Is it the desired behavior ?

components/connect-github.tsx
```diff
@@ -2,29 +2,17 @@ import { useTranslation } from "next-i18next";
...
 
 
-  if(size === 'sm'){
-    return (
-    <Button onClick={connectGithub} disabled={isConnecting} isLoading={isConnecting}> 
-      <GithubImage  opacity={1} /> 
-      <span>{t("actions.connect")}</span>
-    </Button>)
-  }
-
```
clarkjoao, line 27: why this button has been removed?

x-hooks/use-network.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+      dispatch(changeAllowedTokens(transactional, reward));
+    })
+  }
+
+  function loadNetworkTimes() {
+    if (!state.Service?.active || !state?.Service?.network?.active)
+      return;
+
+    const network: any = state.Service.network.active;
```
vhcsilva, line 126: I think we have a conflict here, because `state.Service.network.active` is a database network object and not a Network_v2 object

components/connect-github.tsx
moshmage: @clarkjoao might have been a miss-rebase will match against master before merging so we don't lose actual work :)

x-hooks/use-bounty.tsx
moshmage: @vhcsilva as we saw yesterday, if we do that we set an expensive behaviour on the application; Since these two functions should only be needed _some times_ and not _everytime_ we want a bounty data -- these need to be called by hand, so we make sporadic use of them _and_ control when they are called so we don't end up on making this call too many times.

x-hooks/use-bounty.tsx
moshmage: Is a bit weird that `contractId` can be 0 since the `0` bounty on the contract is an empty bounty :O

x-hooks/use-bounty.tsx
```diff
@@ -117,7 +127,7 @@ export function useBounty() {
     dispatch(changeSpinners.update({bountyChain: true}))
 
     state.Service.active.getBounty(state.currentBounty.data.contractId)
-      .then(bounty => {
+      .then(async bounty => {
```
moshmage, line 130: no asyncs inside a non-async function, use then callback or affect state change so erc20data is loaded on its own side-effect

PR #624, opened 2022-10-18
Dev 893 logos are not being shown on my network page

MarcusviniciusLsantos:


components/profile/my-network-settings.tsx
```diff
@@ -307,6 +308,19 @@ export default function MyNetworkSettings({ network, updateEditingNetwork } : My
     }
   }, [network?.networkAddress]);
 
+  useEffect(() => {
+    const logoSize = details?.fullLogo?.value?.raw?.size
+    const iconSize = details?.iconLogo?.value?.raw?.size
+
+    const megaByte = 1000000
+    if(logoSize >= megaByte || iconSize >= megaByte || logoSize + iconSize >= megaByte){
```
vhcsilva, line 316: I think just the `logoSize + iconSize >= megaByte` condition is enough as if it is `true` one of the other conditions will also be `true`

MarcusviniciusLsantos: in this situation I needed all the conditions, because the image when it comes from the database comes just the way and I don't get its size because it has already been validated at another time. Because of this, there are NaN situations when variables are changed for the 1st time

MarcusviniciusLsantos: yea. and I didn't think about it :( my bad xD

moshmage (approved):

PR #635, opened 2022-10-24
Dev 981 after creating a funding request and the draft time passes it updates the state progress even without the funded it is complete

MarcusviniciusLsantos:


moshmage:
@MarcusviniciusLsantos update board accordingly, please

moshmage (approved):

PR #639, opened 2022-10-26
DEV-911 update alloance when open modal

clarkjoao:


components/create-bounty-modal.tsx
```diff
@@ -611,13 +611,15 @@ export default function CreateBountyModal() {
   }, [customTokens]);
 
   useEffect(() => {
+    let approved = true
```
vhcsilva, line 614: It should be `false` by default, this way we prevent the user from getting the `exceeds allowance` error if the app fails to update the allowance

components/create-bounty-modal.tsx
```diff
@@ -636,6 +638,10 @@ export default function CreateBountyModal() {
     getTokenInfo(activeNetwork.tokens);
   }, [activeNetwork?.tokens, settings]);
 
+  useEffect(()=>{
+    transactionalERC20.updateAllowanceAndBalance();
```
vhcsilva, line 642:
```suggestion
    transactionalERC20.updateAllowanceAndBalance();
    rewardERC20.updateAllowanceAndBalance();
```

vhcsilva (changes requested):

components/create-bounty-modal.tsx
clarkjoao: yeah, i thought the same, but when the type is foundingBounty the approval must be true, therefore i will improve that.

moshmage: I'd add that explanation as a comment and I think it does suffice

moshmage (approved):

vhcsilva (approved):

PR #654, opened 2022-11-01
Dev 671 missing x will be taken from this value when you create the network warning

MarcusviniciusLsantos:


moshmage:
@MarcusviniciusLsantos reduce branch-name please

PR #693, opened 2022-11-14
DEV-1081 check delegate erros

clarkjoao:


components/oracles-delegate.tsx
```diff
@@ -55,8 +56,18 @@ function OraclesDelegate({
...
 
   function handleChangeAddress(params: ChangeEvent<HTMLInputElement>) {
-    if (error) setError("");
+    if(addressError) setAddressError("")
     setDelegatedTo(params.target.value);
+
+    if(Service?.active?.web3Connection && params.target.value){
+      clearTimeout(debounce.current)
+    
+      debounce.current = setTimeout(() => {
```
moshmage, line 65: there has to be a better way than making a setTimeout of half-a-second

clarkjoao: there is just to better the UX, while the user type the address not doing validation and render ther warning in screen.

PR #706, opened 2022-11-17
Dev 1027 show dispute actor on a list on the proposal page

MarcusviniciusLsantos:


components/curator-list-item.tsx
```diff
@@ -0,0 +1,84 @@
...
+        </span>
+      </div>
+
+      <div className="col-2 d-flex flex-row align-items-center pe-4 justify-content-center">
+        <span className="caption-medium text-white">
+          {curator?.disputedProposals || 0}
+        </span>
+      </div>
+
+      <div className="col-2 d-flex flex-row align-items-center justify-content-center">
```
moshmage, line 49: this isn't a row, it's a column -- how can we have both `col` and `flex-row` on this???!

components/network-tx-button.tsx
```diff
@@ -105,8 +107,10 @@ function networkTxButton({
           }));
 
           if (answer.blockNumber)
-            processEvent("oracles","changed", state.Service?.network?.lastVisited, {fromBlock:answer.blockNumber})
-              .catch(console.debug);
+            processEvent("oracles",
```
moshmage, line 110: Like on the previous review, this component should not have such a specific event-processor call since this can be used app-wide and not just for lock/unlock

interfaces/issue-data.ts
```diff
@@ -50,7 +50,14 @@ export interface IssueData {
   working: string[];
   fundedAt?: Date;
   benefactors?: fundingBenefactor[];
-  disputes?: {address: string; weight: number}[];
+  disputes?: disputes[];
+}
+
+export interface disputes {
```
moshmage, line 56: This interface needs to be defined _before_ being used (for readability sake). Also, start the naming with a capital letter, `Disputes` instead of a lower cased one

moshmage (changes requested):

PR #748, opened 2022-12-06
Dev 1151 theres no limitation for title length

MarcusviniciusLsantos:


components/create-bounty-details.tsx
```diff
@@ -15,6 +16,17 @@ export default function CreateBountyDetails({
   review = false,
 }) {
   const { t } = useTranslation("bounty");
+  const [errorTitle, setErrorTitle] = useState<boolean>(false);
```
moshmage, line 19: there's no need to have `setErrorTitle` as we know the length of the title, so we can simply do `title.length > limit ? show-error : null `

moshmage (changes requested):

moshmage (approved):

PR #749, opened 2022-12-06
Dev 1141 theres no limitation for description length

MarcusviniciusLsantos:


components/create-bounty-details.tsx
```diff
@@ -15,6 +20,36 @@ export default function CreateBountyDetails({
...
+        }/${file.hash}) \n\n`);
+
+    const body = `${bountyDescription}\n\n${strFiles
+          .toString()
+          .replace(",![", "![")
+          .replace(",[", "[")}`
+
+    body?.length && setBodyLength(body.length)
+
+    if(body.length > BODY_CHARACTERES_LIMIT && !errorDescription) setErrorDescription(true)
```
moshmage, line 49: this can be done directly on the html as we should know the `body.length` and the `filesLenght`, summing the two would be enough to show an error -- there shouldn't be a need for a `setErrorDescription`

moshmage (changes requested):

moshmage (approved):

PR #751, opened 2022-12-07
DEV-1185 fix update network tokens

clarkjoao:


moshmage (changes requested):
? the only change on the PR is variables names. If that fixed anything, then something wasn't broken.

clarkjoao:
> ? the only change on the PR is variables names. If that fixed anything, then something wasn't broken.

Trust me, it fix! 😅 
The variable name has the same name that function, so the parameters at line 148 was receiving a function and not the array.

moshmage (approved):

PR #752, opened 2022-12-07
DEV-1174 fix registry setup inputs

clarkjoao:


components/setup/registry-setup.tsx
```diff
@@ -122,7 +122,7 @@ export function RegistrySetup({
     handleDeployRegistry( erc20.value,
                           lockAmountForNetworkCreation,
                           treasury,
-                          networkCreationFeePercentage,
+                          `${+networkCreationFeePercentage / 100}`, // TO NORMALIZE INPUT BETWEEN 0 AND 100
```
moshmage, line 125: ?

moshmage (changes requested):

components/setup/registry-setup.tsx
clarkjoao: we have resolved in call.

moshmage (approved):

PR #764, opened 2022-12-14
Dev 1187 new proposal list distribution

clarkjoao:


components/proposal-list-distribution.tsx
```diff
@@ -0,0 +1,103 @@
...
+  distributedAmounts,
+}: Props) {
+  const { t } = useTranslation(["common", "proposal"]);
+
+  const [coinInfo, setCoinInfo] = useState<TokenInfo>();
+
+  const { state } = useAppState();
+
+  async function getCoinInfo() {
+    await getCoinInfoByContract(state.Service?.network?.networkToken?.symbol)
```
moshmage, line 30: what you want is `getCoinPrice(state.Service?.network?.networkToken?.symbol)`

components/proposal-progress.tsx
```diff
@@ -1,40 +1,97 @@
...
+  progressColor = "primary",
+}) {
+  return (
+    <div
+      className={
+        "rounded-bottom d-flex flex-column align-self-end align-items-center gap-2 min-w-fit"
+      }
+      style={{ width: `${percentage}%` }}
+    >
+      {githubLogin ? (
```
moshmage, line 30: guess we shouldn't have deleted that GithubAvatar component heh? [re-introduce that component and use it where necessary in this PR]

moshmage (changes requested):

components/proposal-list-distribution.tsx
clarkjoao: its used on `handleConversion`, to convert the value by coingecko

PR #778, opened 2022-12-20
Dev 1185 fix update network tokens

clarkjoao:


components/profile/my-network-settings/tokens-settings.tsx
```diff
@@ -27,235 +32,152 @@ export default function TokensSettings({
...
-          fromBlock: (txInfo as { blockNumber: number }).blockNumber 
-        })
-        getAllowedTokensContract();
+    if(toAdd.length) transactions.push(state.Service?.active.addAllowedTokens(toAdd, isTransactional))
+    if(toRemove.length) transactions.push(state.Service?.active.removeAllowedTokens(toRemove, isTransactional))
+
+    Promise.all(transactions).then(([txAdd, txRemove])=>{
+      const blockAdd = (txAdd as { blockNumber: number })?.blockNumber 
+      
+      if(!blockAdd) return;
```
vhcsilva, line 111: Did you tested only removing tokens ? I think that due to this condition it won't work, because if we only remove tokens the `blockAdd` will be `undefined` and the function will always return before the updates.

vhcsilva (changes requested):

components/profile/my-network-settings/tokens-settings.tsx
clarkjoao: > Did you tested only removing tokens ? I think that due to this condition it won't work, because if we only remove tokens the `blockAdd` will be `undefined` and the function will always return before the updates.

will not, because have a if adding the transactions at array, so if we just remove one token, the action remove will be the first transaction. (line 104 and 105)


` const transactions = []

    if(toAdd.length) transactions.push(state.Service?.active.addAllowedTokens(toAdd, isTransactional))
    if(toRemove.length) transactions.push(state.Service?.active.removeAllowedTokens(toRemove, isTransactional))

    Promise.all(transactions).then(([txAdd, txRemove])=>{
      const blockAdd = (txAdd as { blockNumber: number })?.blockNumber 
      
      if(!blockAdd) return;`

vhcsilva: As we only need to know the lowest block, maybe we can use a `reduce` on the `Promise.all` result to get it

moshmage: while the explanation is true (since only one index exists at the array, "blockAdd" will in effect be "blockRemove") the naming of the variables will induce in error when re-reading this code later on (hell: I've had a hard time understanding it _now_ and I had this comment explaining xD)

I would _rename_ the `blockAdd` and `blockRemove` to `firstTx` and `secondTx` accordingly, that way when reading the code again the coder will _clearly_ understand what should be happening

clarkjoao: @moshmage @vhcsilva you guys can check the fix here: https://github.com/bepronetwork/web-network/pull/799

PR #779, opened 2022-12-21
DEV-976 fix bounty roadmap

clarkjoao:


x-hooks/use-bounty.tsx
```diff
@@ -152,11 +152,11 @@ export function useBounty() {
         getExtendedProposalsForCurrentBounty(bounty)
           .then(proposals => dispatch(changeCurrentBountyDataProposals(proposals)))
 
-        bountyReadyPRsHasNoInvalidProposals(bounty, state.Service.network)
+        bountyReadyPRsHasNoInvalidProposals(bounty, state.Service.active)
```
vhcsilva, line 155: The second parameter is supposed to be a Network_v2 object, it should be `state.Service.active.network`

vhcsilva (changes requested):

PR #792, opened 2022-12-27
DEV-1135 add spinner when connect gh

clarkjoao:


MarcusviniciusLsantos (approved):

x-hooks/use-authentication.tsx
```diff
@@ -119,12 +119,15 @@ export function useAuthentication() {
...
+          return dispatch(changeConnectingGH(false))
 
         lastUrl.value = asPath;
+        
+        if(signedIn)
+          signIn('github', {callbackUrl: `${URL_BASE}${asPath}`})
 
-        return signedIn ? signIn('github', {callbackUrl: `${URL_BASE}${asPath}`}) : null;
-      }).finally(()=> dispatch(changeConnectingGH(false)))
+        return setTimeout(() => dispatch(changeConnectingGH(false)), 5 * 1000)
```
moshmage, line 129: since we are sending the user to another page when we call `signIn` we don't actually need to dispatch this as the app will reload the state when the user comes back.

if you want to make sure that the state is correct when the function starts again you can simply add the `dispatch(changeConnectingGH(false))` bit of code to the line 105 of this file; that way it doesn't matter what state it is, it will always clean it before starting

clarkjoao: Yes, but sometimes when you already have gh session at your browser, visually you is not redirectly, then, the issue ask for show a spinner while the next auth finish your job.

to make sure the spinner will stop sometime, i return this timeout.

PR #797, opened 2022-12-31
Dev 982 kyc

clarkjoao:


components/page-actions.tsx
```diff
@@ -245,7 +246,14 @@ export default function PageActions({
         !isWorkingOnBounty &&
         isRepoForked &&
         isStateToWorking)
-      return (
+
+      if(state.currentBounty?.data?.isKyc 
+          && state?.currentUser?.kyc
```
moshmage, line 251: you dont need to check for `state?.currentUser?.kyc && state?.currentUser?.kyc?.status` since `?.` (null coalescent) should take care of that for you

pages/api/kyc/init.ts
```diff
@@ -0,0 +1,58 @@
...
+    const user = await models.user.findOne({
+      where: { 
+        githubLogin: session.user.login
+      },
+      raw: true
+    })
+
+    kycSession = await models.kycSession.findOne({
+       where:{
+          user_id: user.id,
```
moshmage, line 27: exit with 400 if user was not found (for some weird reason)

pages/api/kyc/validate.ts
```diff
@@ -0,0 +1,74 @@
...
+      kycSession.save();
+    }
+
+    return res.status(200).json(kycSession);
+  } catch (error) {
+    return res.status(500).json(error);
+  }
+}
+
+export default async function GetIssues(req: NextApiRequest,
```
moshmage, line 62: `GetIssues` does not match name of the file

helpers/query/findUserBySession.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const user = await models.user.findOne({
+      where: { 
+        githubLogin: session.user.login
+      },
+      raw: true
+  })
+
+  if(!user)
+    return res.status(400).json('User not found')
```
moshmage, line 22: return null and deal with what should happen to that on the logic that uses this function

moshmage (changes requested):
rebase on branch needed

PR #804, opened 2023-01-05
Dev 777 bounties discovery page

MarcusviniciusLsantos:


pages/api/search/issues/index.ts
```diff
@@ -150,6 +176,64 @@ async function get(req: NextApiRequest, res: NextApiResponse) {
                                     .map((value)=> value === '+' ? Sequelize.literal('+') : 
                                       (COLS_TO_CAST.includes(value) ? castToDecimal(value) : value));
 
+    if(lastEdited) {
```
moshmage, line 179: code complexity on this endpoint has grown beyond control; Would be simpler to add these conditions to their own route - this way we _don't add more_ technical debt.

It would also help identify repeated block-codes that can be refactored into a helper function to further reduce technical debt on the added code

interfaces/issue-data.ts
```diff
@@ -70,6 +71,18 @@ export interface IssueBigNumberData extends Omit<IssueData , "amount" | "funding
   fundedPercent: BigNumber;
 }
 
+export interface IssueNetwork extends IssueBigNumberData {
+  networkName?: string; 
+  totalValueLock?: BigNumber; 
+  issues?: IssueBigNumberData[]
+}
+export interface IssueSearch {
+  rows: IssueNetwork[] | [],
```
moshmage, line 80: rows _can't_ be `never[]` _it has_ to have a type.

components/issue-list-item.tsx
```diff
@@ -73,94 +98,44 @@ export default function IssueListItem({
...
-          },
-        });
-      }}
-    >
+    <CardItem onClick={handleClickCard}>
       <div className="row align-center">
         <div className="col-md-10 mb-3 mb-md-0">
           <h4 className="h4 text-truncate mb-3">
-            <span className="text-white-40 me-2">#{issue?.githubId}</span>
+            <Tag/>
```
moshmage, line 138: This name will certainly raise conflicts with the tags coming in from the other story, change it to better naming would be `creatorTag()` that way we know _what_ will be inside the tag

interfaces/issue-data.ts
```diff
@@ -55,6 +56,11 @@ export interface IssueData {
   benefactors?: fundingBenefactor[];
   disputes?: Disputes[];
   payments: Payment[];
+  network?: {
+    colors: ThemeColors;
```
moshmage, line 60: _we don't need a complete copy of the theme colours, right?_ we just need the primary?

interfaces/issue-data.ts
```diff
@@ -55,6 +56,11 @@ export interface IssueData {
   benefactors?: fundingBenefactor[];
   disputes?: Disputes[];
   payments: Payment[];
+  network?: {
```
moshmage, line 59: add _information_ to the property name so we understand that's information and not the actual _network_.

pages/api/search/issues/total.ts
```diff
@@ -38,6 +38,17 @@ async function getTotal(req: NextApiRequest, res: NextApiResponse) {
...
     whereCondition.network_id = network?.id;
+  } else {
+    const networks = await models.network.findAll({
+      where: {
+        isRegistered: true,
+        isClosed: false
+      }
+    })
+
+    if (networks.length === 0) return res.status(404).json("Networks not found");
```
moshmage, line 49: answer with a json message instead or switch to `text()`

pages/bounties.tsx
```diff
@@ -0,0 +1,107 @@
+import { useEffect, useState } from "react";
```
moshmage, line 1: change filename to `bounty-hall` to match title

pages/bounties.tsx
```diff
@@ -0,0 +1,107 @@
...
+    {
+      value: 0,
+      label: t("bounty:distributed"),
+      currency: "USD",
+    },
+  ]);
+
+  
+  useEffect(() => {
+    if (state.Service?.active) state.Service?.active.loadRegistry();
```
moshmage, line 50: the registry should already be loaded when we get to this state

x-hooks/use-api.tsx
```diff
@@ -72,7 +73,7 @@ export default function useApi() {
     repoId = "",
     time = "",
     state = "",
-    sortBy = "updatedAt",
+    sortBy = "createdAt",
```
moshmage, line 76: should we change the default, or should we just use `createdAt` as a param when calling this function to that case?

pages/api/search/issues/total.ts
MarcusviniciusLsantos: Is the standard we are following on endpoints

image

helpers/handleNetworksValuesApi.ts
```diff
@@ -0,0 +1,26 @@
+const newNetworkObj = (issue) => ({
```
moshmage, line 1: redundant, code can live inside the `handleNetwork` function as is as it's not re-used,

```js
const handleNetwork(issues) =>
  issues?
    .filter(i => i?.network)
    .map(({network: {name, logoIcon, colors: {primary})) => 
      ({name, logoIcon, colors: {primary}})
```

helpers/handleNetworksValuesApi.ts
```diff
@@ -0,0 +1,26 @@
...
+    return issue;
+  });
+
+export default function handleNetworkValues(issues) {
+  if (issues?.rows) {
+    return {
+      ...issues,
+      rows: handleNetwork(issues.rows),
+    };
+  } else if (issues?.length > 0) {
```
moshmage, line 23: `issues?.lenght` is error prone, `helloworld` has length but it's not an array;

```js
else if (Array.isArray(issues) && issues?.length > 0)
```

pages/api/search/issues/index.ts
```diff
@@ -52,8 +54,22 @@ async function get(req: NextApiRequest, res: NextApiResponse) {
       });
 
       if (!network) return res.status(404).json("Invalid network");
-
+      networks = [network]
       whereCondition.network_id = network?.id;
+    } else {
```
moshmage, line 59: use `if (allNetworks)` instead, as we might fire this else condition in an edge case, and using a strict flag will solve that.

moshmage (changes requested):

x-hooks/use-repos.ts
```diff
@@ -25,7 +25,8 @@ export function useRepos() {
   const {getReposList} = useApi();
   const { getRepository, getRepositoryForks, getRepositoryBranches, getRepositoryViewerPermission } = useOctokit();
 
-  function loadRepos(force = false, name = state?.Service?.network?.lastVisited) {
+  function loadRepos(force = false) {
```
moshmage, line 28: make sure no function was calling this with a second argument

moshmage (approved):

PR #806, opened 2023-01-09
Dev 778 bounties leaderboard

vhcsilva:


pages/api/search/leaderboard/points/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    const leaderboard = await models.leaderBoard.findAndCountAll(paginate({
+      attributes: {
+        exclude: ["id"],
+      },
+      where: whereCondition,
+      nest: true,
+      raw: true
+    }, req.query, [[sortBy || "numberNfts", order || "DESC"]]))
+      .then(async (items) => {
+        return Promise.all(items.rows.map(async (item) => {
```
moshmage, line 27: if all you need is the addresses of the users, you can join them all and then make one big request with
```ts
model.user.findAll({where: { address: { [Op.in]: items.rows.map(({address}) => address.toLowerCase()) } }})
```

You can easily separate the `leaderboard` constant and this _then_ into its own constants; this way you can then access the leaderboard prop to reduce/map the result found on the batch-user findAll.

scripts/leaderboard.js
```diff
@@ -0,0 +1,146 @@
+require("dotenv").config();
+
```
moshmage, line 2: I understand this is a javascript version of the file above, same review applies.

With the added question: _can't we re-use code?!_

moshmage (changes requested):

moshmage (approved):

PR #808, opened 2023-01-10
Dev 693 add bounty tags and enable bounty creators to associate their bounty with tags

vhcsilva:


moshmage:
@vhcsilva I think this branch name will be to big to be built on jenkins

vhcsilva:
@moshmage it was already deployed on jenkins, it worked

moshmage (approved):

PR #812, opened 2023-01-12
Fix/add migrate to create get transfer events database

MarcusviniciusLsantos:


moshmage:
@MarcusviniciusLsantos please rebase your branch, you only have one change so you should only have one commit on this PullRequest

db/migrations/2022120207400001-add-transfer-events-to-chain-events-table.js
```diff
@@ -0,0 +1,47 @@
...
+
+const {
+  Web3Connection,
+} = require("@taikai/dappkit");
+require("dotenv").config();
+
+const name = "add-transfer-events-to-chain-events-table";
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
```
moshmage, line 14: PS: Make sure that the event-name does not yet exist before adding it to the list

PR #813, opened 2023-01-13
DEV 946 rss feed of bounties

vhcsilva:


templates/rss.ts
```diff
@@ -0,0 +1,33 @@
+export const rssTemplate = `<?xml version="1.0" encoding="UTF-8" ?>
```
moshmage, line 1: wouldn't it make sense for this file to be `tsx` ?

db/migrations/20230113214456-add-rss-time-to-settings.js
```diff
@@ -0,0 +1,20 @@
...
+    await Settings.create({
+      key: "rssTtl",
+      value: "1",
+      visibility: "public",
+      type: "string"
+    });
+  },
+
+  async down (queryInterface, Sequelize) {
+
```
moshmage, line 18: remove the changes made on `up`

templates/rss.ts
vhcsilva: @moshmage as this file only exports a string and not HTML, i think we can use the `.ts` extension

moshmage: if this is a template for handlebars, use `hbs` instead; If this is "html with javascript" use `tsx`

moshmage (approved):

PR #822, opened 2023-01-17
Bepro 1318 change curators structure behaviour

MarcusviniciusLsantos:


contexts/network-settings.tsx
```diff
@@ -209,7 +209,7 @@ export const NetworkSettingsProvider = ({ children }) => {
       validator: async (value: string) => {
         let validated = undefined;
   
-        if (value.trim() !== "")
+        if (value.trim() !== "" && !isSetup)
```
moshmage, line 212: this condition was removed on development, rebase or merge dev into your branch to make sure you have the latest changes :)

moshmage (changes requested):

pages/[network]/curators/index.tsx
```diff
@@ -1,18 +1,64 @@
...
 export default function PageCouncil() {
-  return <ReadyToPropose/>
+  const { t } = useTranslation(["council"]);
+  const router = useRouter();
+  const { type } = router.query;
+
+  const types = {
+    "curators-list": <CuratorsList key={"curators-list"} />,
+    "ready-to-close": (
+      <ListIssues
```
Comment (moshmage) [inline pages/[network]/curators/index.tsx:20]: add a `inView: boolean` prop to this component that _will not_ render or start useEffects if it's not true -- as long as it exists:

```
// on this object, add
inView={type === "ready-to-close"}


// on ListIssues component, add the following pseudo-logic
{inView: null}
// ...
if (inView !== null && inView === false)
  return;

**Review by moshmage — CHANGES_REQUESTED** [review]

File: components/curators-list.tsx [hunk 3]
```diff
@@ -39,7 +39,7 @@ export default function CuratorsList() {
   const { state, dispatch } = useAppState();
 
   function handlerSearch() {
-    if (!state.Service?.network?.active) return;
+    if (!state.Service?.network?.active && inView === false) return;
```
vhcsilva, line 42: I think that the correct condition is `!state.Service?.network?.active || inView === false`, because when `state.Service?.network?.active` is `true` and `inView` is `false` it won't return

components/list-issues.tsx
```diff
@@ -163,7 +165,7 @@ export default function ListIssues({
   }
 
   function handlerSearch() {
-    if (!appState.Service?.network?.active?.name) return;
+    if (!appState.Service?.network?.active?.name && inView === false) return;
```
vhcsilva, line 168: Same as above condition

components/list-issues.tsx
```diff
@@ -67,7 +68,8 @@ export default function ListIssues({
   proposer,
   redirect,
   disputableFilter,
-  allNetworks = false
+  allNetworks = false,
+  inView
```
vhcsilva, line 72: This component is used by other pages too, such as `/profile/bounties`, `/profile/pull-request`, `/profile/proposals`, etc. Have you checked that everything is fine with these pages ?

MarcusviniciusLsantos: yes, they all rendered normally

vhcsilva (approved):

moshmage (approved):

PR #831, opened 2023-01-20
Bepro 1323 change table of tokens to have column is reward so we dont have dupes on database

MarcusviniciusLsantos:


db/migrations/2023011917330000-add-column-isReward-to-tokens-table.js
```diff
@@ -0,0 +1,28 @@
...
+      allowNull: true,
+      defaultValue: false,
+    });
+
+    Tokens.init(queryInterface.sequelize);
+//mesmo address mesmo column?
+    const tokens = await Tokens.findAll();
+    for (const token of tokens) {
+      console.log('token ->', token.isReward, token.isTransactional)
+      if (token.isTransactional === true) token.isReward = false;
```
moshmage, line 18: can be changed to `token.isReward = token.isTransactional === false;`

db/migrations/2023011917330000-add-column-isReward-to-tokens-table.js
```diff
@@ -0,0 +1,28 @@
...
+      else token.isReward = true;
+
```
moshmage, line 20: if `token` is reward and we changed the state, we should now have duplicates and these need to be removed

moshmage (changes requested):

db/migrations/2023011917330000-add-column-isReward-to-tokens-table.js
vhcsilva: @moshmage @MarcusviniciusLsantos i think that this migration instead of only set `true` or `false` to `isReward`, should iterate over tokens and check if the same token is allowed as transactional/reward, update `isTransactional` and `isReward` columns according to it and then remove the duplicated records.

MarcusviniciusLsantos: I adjust the migration to add in what it needs and delete the duplicate. I'll just need to adjust the rest to see if the logics are working correctly that way. because now IsAllowed is false when isTransactional and isReward is false

components/create-bounty-modal.tsx
```diff
@@ -172,7 +172,7 @@ export default function CreateBountyModal() {
         decimals: transactionalERC20?.decimals,
         amount: rewardAmount,
         setAmount: setRewardAmount,
-        tokens: customTokens.filter(token => token?.isTransactional !== true),
+        tokens: customTokens.filter(token => token?.isReward !== false),
```
moshmage, line 175: isn't it the same thing?

components/profile/my-network-settings/tokens-settings.tsx
```diff
@@ -58,11 +58,11 @@ export default function TokensSettings({
         dbTransactionalAllowed
       } = dbTokens.reduce((previous, current) => {
         const tmp = { ...previous };
-        const { isTransactional, isAllowed } = current;
+        const { isTransactional, isAllowed, isReward } = current;
```
moshmage, line 61: this code can be simplified now that we have the `isReward`, something like this:

```ts
const dbTokens = await getTokens();
const availableRewardList = [];
const availableTransactionList = [];

for (const token of dbTokens) {
  if (token.isReward)
    availableRewardList.push(token);
  else if (token.isTransactional)
    availableTransactionList.push(token);
}

if (isGovernorRegistry)  {
  /** unsure if this is needed */
  setSelectedRewardTokens(availableRewardList);
  setSelectedTransactionalTokens(availableTransactionList);
  /** but kept it because previous code had it. */

  availableRewardList.push(...availableTransactionList);
  availableTransactionList.push(...availableRewardList)
}

setAllowedRewardTokensList(availableRewardList)
setAllowedTransactionalTokensList(availableTransactionList)


```

moshmage (changes requested):

components/create-bounty-modal.tsx
MarcusviniciusLsantos: if there is no allowed in the registry with this condition, it can take any token that exists in our database.

components/profile/my-network-settings/tokens-settings.tsx
MarcusviniciusLsantos: I tried to apply this logic, but there were some use cases it didn't cover. how we use this component for governance and registry. I chose to keep and adjust only what was necessary

components/create-bounty-modal.tsx
vhcsilva: It could be simply `token?.isReward` or a double negation `!!token?.isReward`, as we want to know if it is `true`

helpers/handleAllowedTokens.ts
```diff
@@ -17,7 +17,7 @@ export function handleAllowedTokensDatabase(allowedTokens: {
     reward: allowedTokens.reward
       ?.map((rewardToken) => {
         return tokens.find((token) =>
-            token.address === rewardToken && token.isTransactional === false);
+            token.address === rewardToken && token.isReward === true);
```
vhcsilva, line 20: As I wrote before, it can be simplified

components/profile/my-network-settings/governance-settings.tsx
```diff
@@ -89,6 +90,18 @@ export default function GovernanceSettings({
       });
   }
 
+  function handleNetworkTokens(tokens: Token[]) {
+    return tokens.map(token => ({
+        ...token,
+        isReward: token.network_tokens.isReward || false,
+        isTransactional: token.network_tokens.isTransactional || false
```
vhcsilva, line 97: Using double negation will result in `false` when `token.network_tokens.isReward` is `null`, so you can use `!!token.network_tokens.isReward` instead

scripts/settings/save-from-env.js
```diff
@@ -116,7 +117,13 @@ const updateTokens = async ({
...
       where: {
         tokenId: token.id,
-        networkId: beproNetwork.id
+        networkId: beproNetwork.id,
+      },
+      defaults: {
+        tokenId: token.id,
+        networkId: beproNetwork.id,
+        isTransactional: true,
+        isReward: false
```
vhcsilva, line 126: `updateTokens` has a `isTransactional` prop, we can use it to define when it is reward or not, and when the relation already exists we need to update `isTransactional` and `isReward` according to this prop

components/create-bounty-modal.tsx
```diff
@@ -605,15 +605,15 @@ export default function CreateBountyModal() {
   }, [transactionalERC20.allowance, rewardERC20.allowance, issueAmount, rewardAmount, rewardChecked]);
 
   useEffect(() => {
-    if (!Service?.network?.tokens || !showCreateBounty)
+    if (!Service?.network.active.tokens || !showCreateBounty)
```
vhcsilva, line 608: use `Service?.network?.active?.tokens`

vhcsilva (approved):

pages/api/network/index.ts
```diff
@@ -195,19 +195,45 @@ async function post(req: NextApiRequest, res: NextApiResponse) {
     //TODO: move tokens logic to new endpoint   
     if(allowedTokens?.allowedTransactions?.length > 0){
       for (const token of allowedTokens.allowedTransactions) {
-        await Database.networkTokens.create({
-          networkId: network.id,
-          tokenId: token.id
+        const [networkToken, created]= await Database.networkTokens.findOrCreate({
```
moshmage, line 198: This file has enough repeated code that it becomes a pain whenever we need to change something (as seen on this PR).

_I would_ move the "loop, find and create" logic to a closure (or helper) function and then call it from here with the needed parameters. Effectively reducing the complexity and work we need to do when this happens;

moshmage (changes requested):
Too much repeated code at `pages/api/network/index.ts`

pages/api/network/index.ts
MarcusviniciusLsantos: initially that's what I did but I had some problems I'll try again and see if the result holds

moshmage (approved):

PR #849, opened 2023-01-30
Bepro 1319 change network hero to calculate values from the api instead of querying chain

MarcusviniciusLsantos:


pages/api/header/networks/index.ts
```diff
@@ -0,0 +1,43 @@
...
+      state: {[Op.not]: "pending"}
+    }
+  });
+
+  const numberNetworks = await models.network.count({});
+
+  const headerInformation = await models.headerInformation.findAll({})
+
+  if(!headerInformation)
+    return res.status(404).json("Header information not found");
```
moshmage, line 21: use `.json({message: "string"});` so json method is respected (or use `.text()` instead of `json()`)

moshmage (changes requested):

MarcusviniciusLsantos:
> Missing calls that make the event run whenever a user opens a bounty or locks/unlocks tokens on a network so values are updated

we talked about not using it. In order not to fall into time-out. But I can follow and implement this logic inside events

moshmage (approved):

PR #859, opened 2023-02-06
Bepro 1353 change network list items to get data from database instead of blockchain

MarcusviniciusLsantos:


pages/api/search/networks/index.ts
```diff
@@ -29,24 +30,43 @@ async function get(req: NextApiRequest, res: NextApiResponse) {
...
+    { association: "curators", required: false },
+    { association: "issues", required: false,
+      where: { 
+        state: {[Op.not]: "pending" }
+      }
+    }
   ];
 
-  const networks = await models.network.findAndCountAll(paginate({
+  const networks = await models.network.findAll({
```
moshmage, line 42: you can use sequelize to parse the data for you,

```
await models.network.findAll({
  attributes: {
    [
      [sequelize.fn('sum', sequelize.col('curators.tokensLocked')), 'totalValueLock'],
      [sequelize.fn('count', sequelize.col('issues.id')), 'countIssues'],
    ]
  },
  include: [
    ...include,
    {
      model: models.issues,
      as: 'issues',
      where: { state: 'open' },
      attributes: [[sequelize.fn('count', sequelize.col('issues.id')), 'countOpenIssues']]
    }
  ],
  group: ['network.id']
  // can add raw: true if you want
})

**Review by moshmage — CHANGES_REQUESTED** [review]

File: pages/api/search/networks/index.ts (hunk 1 above)
Comment (MarcusviniciusLsantos) [reply pages/api/search/networks/index.ts:76]: For the record, we chose to keep it because the sequelize count for several associations is complex, requiring more than one request to return the expected value.

Comment (moshmage) [reply pages/api/search/networks/index.ts:76]: > For the record, we chose to keep it because the sequelize count for several associations is complex, requiring more than one request to return the expected value.

Will need an example for this affirmation since we have all the needed relationships between tables to make this SQL statement exist and thus Sequelize should also be able to create this statement without needing more than one request

**Review by moshmage — COMMENTED** [review]
Need more information on the "Sequelize is not able to do what we want" part

File: pages/api/search/networks/index.ts (hunk 1 above)
Comment (moshmage) [reply pages/api/search/networks/index.ts:76]: There is a cleaner way of doing this, _but this works_:

```ts
  const result = await Database.network.findAll({
    include: [
      {association: 'curators', attributes: {exclude: ['address', 'id', 'disputedProposals', 'acceptedProposals', 'networkId', 'createdAt', 'updatedAt', 'tokensLocked', 'isCurrentlyCurator']}},
      {association: 'issues', attributes: {exclude: issues_exclude}},
    ],
    attributes: {
      include: [
        [Sequelize.fn('sum', Sequelize.cast(Sequelize.col('curators.tokensLocked'), 'int')), 'tokensLocked'],
        [Sequelize.fn('count', Sequelize.where(Sequelize.col(`issues.state`), {[Op.not]: 'pending'})), 'totalIssues'],
        [Sequelize.fn('count', Sequelize.where(Sequelize.col('issues.state'), {[Op.eq]: 'open'})), 'openIssues'],
      ],
      exclude: ['creatorAddress', 'colors', 'description', 'logoIcon', 'fullLogo', 'isRegistered', 'isClosed', 'councilMembers', 'createdAt', 'updatedAt', 'isDefault'],
    },
    group: ['network.id', 'curators.id', 'issues.id'],
    where: {
      isRegistered: true,
      isClosed: false,
    },

  });
```

proof:
image

moshmage (changes requested):

moshmage (changes requested):
@MarcusviniciusLsantos a thread was made on discord

moshmage (approved):

PR #863, opened 2023-02-08
Bepro 1342 after connecting GitHub in setup page user is redirected to networks page

MarcusviniciusLsantos:


pages/connect-account.tsx
```diff
@@ -56,8 +56,10 @@ export default function ConnectAccount() {
   function redirectToProfile() {
     const lastNetworkVisited = new WinStorage(`lastNetworkVisited`, 0, 'localStorage');
 
+    const toNetworks = state.Service?.active?.network ? "/networks" : "/setup"
```
moshmage, line 59: can't we simply _grab the current page_ and pass that instead?

moshmage (changes requested):

pages/connect-account.tsx
MarcusviniciusLsantos: is the current page the connect-account, in this case would it be the previous page?

moshmage (approved):

moshmage:
@MarcusviniciusLsantos make double-sure that router.back() returns what we actually want

PR #882, opened 2023-02-27
Bepro 1414 is possible to config fees to 100

vhcsilva:


components/profile/my-network-settings/registry-settings.tsx
```diff
@@ -13,29 +13,88 @@ import { WarningSpan } from "components/warning-span";
...
 
+  function validateLimits(param: RegistryParameters, value: string | number) {
+    if (ParameterValidator(param, value)) return undefined;
+
+    const { min, max } = REGISTRY_LIMITS[param] || {};
+
+    if (min !== undefined && max !== undefined)
+      return t("setup:registry.errors.exceeds-limit", { min, max });
+    else
+      return t("setup:registry.errors.greater-than", { min });
```
moshmage, line 45: if min is undefined, will it show "needs to be greater than 0"?

vhcsilva: No, but I just sent a commit refactoring this function

moshmage (approved):

PR #896, opened 2023-03-02
Bepro 1370 fix UI bounty hall logos main nav logo and bounty hero states

MarcusviniciusLsantos:


components/bounty/bounty-tags.tsx
```diff
@@ -15,8 +17,9 @@ export default function BountyTags({
         <Badge
           key={tag}
           label={tag} 
-          className="caption-small border border-primary border-radius-8" 
-          color="primary-30"
+          className="caption-small border-radius-8" 
+          color="primary" 
+          style={color ? { backgroundColor: color }: null}
```
vhcsilva, line 22: In some cases these badges have transparency in the background color, it still happens ? You can do it by adding the transparency value in the color hex code. Example: https://github.com/bepronetwork/web-network/blob/DEV-961/components/chain-badge.tsx#L15

And the border color needs to change too. Example: 
Color to apply : `#17E708`
Border color: `#17E708`
Background color: `#17E70890`

vhcsilva (changes requested):

components/bounty/bounty-tags.tsx
MarcusviniciusLsantos: I added this new logic and put an opacity boolean because background-30 is caching the bepro color by default

vhcsilva (approved):

PR #911, opened 2023-03-09
Bepro 1452 mc if wallet is already in use and user has a gh session that does not match modal does not show warning the user of that

MarcusviniciusLsantos:


x-hooks/use-authentication.tsx
```diff
@@ -152,6 +154,10 @@ export function useAuthentication() {
           push(`/connect-account`);
           return false;
         }
+        if(user?.githubLogin){
```
moshmage, line 157: this shouldn't be needed, look for `changeCurrentUserMatch` and check why it's not working properly

PR #919, opened 2023-03-16
Bepro 1501 mc make web network use new events api

vhcsilva:


moshmage:
please,, rebase something; it will b impossible to review this PR otherwise :(

PR #927, opened 2023-03-20
Bepro 1361 create bounty form refactor

MarcusviniciusLsantos:


components/create-bounty/create-bounty-network-dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  async function selectSupportedNetwork({ value }) {
+    const network = networks.find((network) => network.name === value);
+
+    onSelect(network);
+    setSelected(networkOption(network));
+  }
+
+  function updateOptions() {
+    if (networks.length > 0) {
+      setOptions(networks.map((network) => ({
```
moshmage, line 73: this map can call `networkOption` function

components/create-bounty/create-bounty-token-amount.tsx
```diff
@@ -0,0 +1,186 @@
...
+        disabled={review}
+        defaultToken={defaultToken}
+        showCurrencyValue={false}
+        needsBalance={isFunding}
+        noLabel
+      />
+    );
+  }
+
+  function inputNumber() {
```
moshmage, line 95: inputNumber is not re-used, we gain nothing but complication by having it as a function

styles/styles.scss
```diff
@@ -78,6 +78,14 @@ li {
   border-color: var(--bs-purple);
 }
 
+.border-dotted {
```
moshmage, line 81: border-dotted and -none can exist together,

 ```
border {
&-dotter {}
&-none {}
}
```

File: x-hooks/use-dao.tsx [hunk 4]
```diff
@@ -64,19 +64,23 @@ export function useDao() {
...
       return;
 
     if (state.Service?.active?.network?.contractAddress === networkAddress)
       return;
 
     const service = state.Service.active;
 
-    if (+chain_id !== +chain.chainId || chain.chainRpc !== state.Service?.active?.web3Host)
-      return;
+    if(chain){
```
moshmage, line 79: ?

MarcusviniciusLsantos: it was necessary because when directly accessing "/create-bounty" it loaded the first chain, but when changing to another chain it did not allow me to load the new DAO.

pages/create-bounty.tsx
```diff
@@ -0,0 +1,768 @@
...
+    setCustomTokens(currentNetwork?.tokens);
+  }, [currentNetwork, connectedChain]);
+
+  useEffect(() => {
+    cleanFields();
+    transactionalERC20.updateAllowanceAndBalance();
+    rewardERC20.updateAllowanceAndBalance();
+  }, []);
+
+  async function handleNetworkSelected(chain: SupportedChainData) {
```
moshmage, line 488: Like previously said, this function shouldn't be calling connect - instead it should instruct the connect to happen from the dropdown

moshmage (changes requested):

moshmage (approved):

PR #941, opened 2023-03-29
Bepro 1495 support multiple networks on profile wallet page

MarcusviniciusLsantos:


services/coingecko.ts
```diff
@@ -56,13 +56,13 @@ const getCoinInfoByContract = async (search: string): Promise<TokenInfo> => {
   if (!publicRuntimeConfig.enableCoinGecko)
     return {prices: {}} as any; // eslint-disable-line
 
-  const storage = new WinStorage(`coingecko:${search.toLowerCase()}`);
+  const storage = new WinStorage(`coingecko:${search?.toLowerCase()}`);
```
moshmage, line 59: `search` is a mandatory argument, if it's null there's some bug on the thing calling this function. Either that, or the "if value" should be made _on the function that calls this one_.

moshmage (changes requested):

components/profile/wallet-balance.tsx
```diff
@@ -57,64 +72,101 @@ export default function WalletBalance() {
     };
   }
 
+  async function handleNetworkSelected(chain: SupportedChainData) {
```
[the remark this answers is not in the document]
moshmage: @MarcusviniciusLsantos this needs to be addressed

services/coingecko.ts
```diff
@@ -90,12 +90,12 @@ async function getCoinPrice(search: string, fiat = settings?.currency.defaultFia
   if (!publicRuntimeConfig.enableCoinGecko)
     return 0;
 
-  const storage = new WinStorage(`coingecko:${search.toLowerCase()}`);
+  const storage = new WinStorage(`coingecko:${search?.toLowerCase()}`);
```
moshmage, line 93: again: this parameter is set as mandatory, if the function calling this does not have a "search" value, then there's a bug on that function. this function shouldn't have a `? ` on search

moshmage (approved):

PR #942, opened 2023-03-29
Bepro 1516 create sqlz migration for existing production database

vhcsilva:


db/migrations/2023010913000001-multichain-updates.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const nullWhereDefaultChainId = [{
+    chain_id: null
+  }, {
+    where: {
+      chain_id: defaultChainId
+    }
+  }];
+
+  await NetworkModel.update(...nullWhereDefaultChainId);
```
moshmage, line 79: I think you could just _not_ have an array and then you wouldn't need to `...` destroy it into an objecto

moshmage (approved):

moshmage:
closing in favour of #957

PR #983, opened 2023-04-10
Bepro 1459 hide bounty from listing

MarcusviniciusLsantos:


vhcsilva (commented):
I think that instead of passing visible: true every time we make a request, we could set visible: true as the default on all endpoints. In this way we guarantee that if we forget to pass this parameter, no hidden bounties will be returned.

vhcsilva (approved):

PR #988, opened 2023-04-12
Bepro 1557 my netwok page new management tab design

MarcusviniciusLsantos:


components/list-issues.tsx
```diff
@@ -92,9 +93,13 @@ export default function ListIssues({
 
   const [searchState, setSearchState] = useState(search);
 
+  const isManagement = variant === 'management';
   const isProfile = variant === "profile";
   const isBountyHall = variant === "bounty-hall";
   const isOnNetwork = !!router?.query?.network;
+  const variantIssueItem = isManagement ? variant : (isProfile || isBountyHall) ? "multi-network" : "network"
+  const columns = ["Name", "Bounty Link", "Hide Bounty", "Cancel"];
```
vhcsilva, line 101: Missing translation

components/profile/my-network-settings/index.tsx
```diff
@@ -291,6 +293,15 @@ export default function MyNetworkSettings({
           </NetworkContainer>
         ),
       },
+      {
+        eventKey: "management",
+        title: "Management",
```
vhcsilva, line 298: Missing translation

pages/api/network/management/index.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (!issue) return res.status(404).json({message: "Issue not found"});
+
+    issue.visible = visible
+
+    await issue.save()
+
+
+    return resJsonMessage("Bounty updated", res, 200);
+  } catch (error) {
+    console.log(error);
```
vhcsilva, line 71: use `Logger` service

vhcsilva (changes requested):
I think that instead of passing `visible: true` every time we make a request, we could set `visible: true` as the default on all endpoints. In this way we guarantee that if we forget to pass this parameter, no hidden bounties will be returned.

vhcsilva (approved):

moshmage (approved):

PR #990, opened 2023-04-13
Bepro 1559 hard cancel bounties on management tab

MarcusviniciusLsantos:


db/migrations/2023041410020002-issues-contractCreationDate-updates.js
```diff
@@ -0,0 +1,102 @@
...
+        const networkV2 = new Network_v2(web3Connection, networkAddress);
+
+        await networkV2.loadContract();
+
+        for (const issue of issues) {
+          const bounty = await networkV2.getBounty(issue?.contractId);
+
+          if (!bounty?.creationDate && !issue?.contractCreationDate) continue;
+
+          const dbBounty = await IssueModel.findOne({
```
vhcsilva, line 86: You already have the db record coming from the first `findAll`, I think you could use it instead of getting it again

vhcsilva (approved):

PR #1003, opened 2023-04-18
Bepro 1400 add note near repository and branch selection on create bounty

MarcusviniciusLsantos:


components/branchs-dropdown.tsx
```diff
@@ -60,11 +72,21 @@ export default function BranchsDropdown({
...
+        options={options?.sort(sortOptions)}
         value={option}
         onChange={onChangeSelect}
         placeholder={t("forms.select-placeholder")}
       />
+      {!option && (
+        <div className="mt-2">
+          <p className="text-gray">
+            {t("support-text-branch")}{" "}
+            <a href="https://discord.gg/layerx" target="_blank">
```
moshmage, line 84: I think we have this value on a variable somewhere, if its not: add it to a constant please

moshmage (approved):

components/branchs-dropdown.tsx
MarcusviniciusLsantos: I also created and adjusted constant from other links

